PR #4, opened 2022-03-30
terraform: added reading from config file capabilities

jabia99:
terraformConfigurator only worked by providing the instances and regions
manually. To make the module more independent, terraformConfigurator can
now read all the necessary information from resources.json


narmaku:
I propose to use newspaper reading code structure:
- https://kentcdodds.com/blog/newspaper-code-structure

Let's discuss it in an ad-hoc meeting!

cloud/terraform/terraform.py
```diff
@@ -123,10 +158,10 @@ def destroy_infra(self):
 
 if __name__ == '__main__':
     tf_conf = TerraformConfigurator('aws')
-    tf_controller = TerraformerController('aws')
+    tf_controller = TerraformController('aws')
 
-    tf_conf.add_region_conf('us-east-1')
-    tf_conf.add_instance_conf('us-east-1', 'ami-0767af0854a146e3e')
+    tf_conf.configure_from_resources_json('resources.json')
```
narmaku, line 163: I would create a dir in the project called "example", "sample" or similar, where we put a real example of resources.json, so that it can be launched to locally test this Terraform controller and configurator. Then we can refer to:
```suggestion
    tf_conf.configure_from_resources_json('sample/resources.json')
```

narmaku (commented):
Nice! Some picky comments added.

cloud/terraform/terraform.py
```diff
@@ -1,12 +1,15 @@
...
 import json
-
+import sys
 
 class TerraformConfigurator:
     def __init__(self, cloud):
         self.cloud = cloud
 
+        # Change cwd to file path
+        os.chdir(sys.path[0])
```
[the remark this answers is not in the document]
narmaku: Good point, creating `main.tf` programmatically would be a good option. Otherwise, we will have to copy the `main.tf` default file to the place where `providers.tf` and `resources.tf` will be created. Maybe it's cleaner to create `main.tf` programmatically. This taking into account that we remove the `.tf` files as part of the "destroy" stage.

cloud/terraform/terraform.py
```diff
@@ -45,39 +44,45 @@ def __new_aws_provider(self, region):
...
-            '\n------------\nProviders:',
-            str(self.providers),
-        )
+        pprint(self.resources)
+        pprint(self.providers)
 
     def set_configuration(self):
+        main_tf = {
+            'terraform': {
+                'required_providers': {'aws': {'source': 'hashicorp/aws', 'version': '~> 3.27'}},
```
narmaku, line 81: I guess we can decouple this part from here and add it programmatically, when we introduce a new cloud.

Or we can add the logic now to set the provider depending on `self.cloud`.

What do you think?

cloud/terraform/sample/resources.json
```diff
@@ -0,0 +1,12 @@
+{
+    "provider": "aws",
+    "instances": [
+        {
+            "ami": "ami-0767af0854a146e3e",
+            "region": "us-east-1",
+            "instance_type": "",
+            "username": "ec2-user",
+            "name": "example-name-RHEL-8.6"
```
narmaku, line 9: Call me picky, but it sounds more natural "sample-name". If not, it would have to be "name-example" I think.

cloud/terraform/terraform.py
```diff
@@ -45,39 +44,45 @@ def __new_aws_provider(self, region):
...
+            self.__new_aws_instance(instance)
+
+    def __new_aws_instance(self, instance):
+        if not instance['instance_type']:
+            instance['instance_type'] = 't2.micro'
+
+        name = instance['name'].replace('.', '-')
 
-    def add_instance_conf(self, region, image_id, name, instance_type=''):
+        # TODO: agree on a name for 'region' or 'location' across all cloud so we can move
```
narmaku, line 59: I think we can keep it like this here, since at this point we are handling AWS-related instances data.

If "region" as such doesn't exist for other clouds, we will handle that in the `__new_<cloud>_instance()` methods, like we do here.

cloud/terraform/terraform.py
```diff
@@ -96,12 +101,7 @@ def configure_from_resources_json(self, resources_path):
     def configure_aws_resources(self, resource_file):
         for instance in resource_file['instances']:
             self.add_region_conf(instance['region'])
-            self.add_instance_conf(
-                instance['region'],
-                instance['ami'],
-                instance['name'].replace('.', '-'),
-                instance['instance_type'],
-            )
+            self.add_instance_conf(instance)
```
narmaku, line 104: Nice!

cloud/terraform/terraform.py
```diff
@@ -45,39 +44,45 @@ def __new_aws_provider(self, region):
...
-            '------------\nResources:',
-            str(self.resources),
-            '\n------------\nProviders:',
-            str(self.providers),
-        )
+        pprint(self.resources)
+        pprint(self.providers)
 
     def set_configuration(self):
+        main_tf = {
```
narmaku, line 79: I would extract this to a separate method. It makes sense to have it decoupled so we can play with more logic there (for multi-clouds).

Then we can simply do (for example) `self.__dump_to_json(self.add_main_conf(), 'main.tf.json')`
(and inside `add_main_conf` we can do the logic to add the corresponding provider to the dict)

narmaku (changes requested):
Small changes proposed, we can sync in a quick call to resolve/discard them.

PR #5, opened 2022-04-01
import ssh key pairs to aws via terraform

jabia99:
to execute tests we need access to the generated instances via ssh.
Added functionality so new resources of type `aws_key_pair` are created,
one per region. Then, each instance get's this key added.

### DIFF
--- a/cloud/terraform/terraform_configurator.py
+++ b/cloud/terraform/terraform_configurator.py
@@ -20,14 +20,15 @@ class TerraformConfigurator:
             }
 
     def configure_from_resources_json(self, resources_path):
+        self.resources_path = resources_path
         with open(resources_path) as f:
             resources_file = json.load(f)
 
         if resources_file['provider'] == 'aws':
             self.configure_aws_resources(resources_file)
 
-    def configure_aws_resources(self, resource_file):
-        for instance in resource_file['instances']:
+    def configure_aws_resources(self, resources_file):
+        for instance in resources_file['instances']:
             self.add_region_conf(instance['region'])
             self.add_instance_conf(instance)
             self.add_ssh_key_conf(instance['region'])
@@ -115,3 +116,15 @@ class TerraformConfigurator:
 
     def set_aws_profile(self, profile):
         self.aws_profile = profile
+
+    def get_username_by_instance_name(self, name):
+        with open(self.resources_path) as f:
+            resources_file = json.load(f)
+
+        for instance in resources_file['instances']:
+            if instance['name'].replace('.', '-') == name:
+                return instance['username']
+            
+        print(f'ERROR: No instance with name "{name}" was found')
+        exit(1)
+        
--- a/cloud/terraform/terraform_configurator.py
+++ b/cloud/terraform/terraform_configurator.py
@@ -4,9 +4,10 @@ from pprint import pprint
 
 
 class TerraformConfigurator:
-    def __init__(self, cloud, ssh_key_path):
+    def __init__(self, cloud, ssh_key_path, resources_path):
         self.cloud = cloud
         self.ssh_key_path = ssh_key_path
+        self.resources_path = resources_path
 
         self.main_tf = {'terraform': {'required_version': '>= 0.14.9'}}
         self.resources_tf = {'resource': {}}
@@ -19,9 +20,8 @@ class TerraformConfigurator:
                 'aws': {'source': 'hashicorp/aws', 'version': '~> 3.27'}
             }
 
-    def configure_from_resources_json(self, resources_path):
-        self.resources_path = resources_path
-        with open(resources_path) as f:
+    def configure_from_resources_json(self):
+        with open(self.resources_path) as f:
             resources_file = json.load(f)
 
         if resources_file['provider'] == 'aws':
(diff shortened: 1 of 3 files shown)


jabia99:
To test this new functionality, I used the AMIs on `resources.json` and checked the connection to each instance via ssh.

cloud/terraform/terraform_controller.py
```diff
@@ -23,12 +24,14 @@ def get_instances(self):
...
         json_output = json.loads(output)
 
-        resources = json_output['values']['root_module']['resources_tf']
+        resources = json_output['values']['root_module']['resources']
 
         instances_info = {}
         if self.cloud == 'aws':
             # 'address' corresponds to the tf resource id
             for resource in resources:
+                if 'aws_instance' not in resource['address']:
```
narmaku, line 35: Q: Shouldn't it be without the `not`?

cloud/terraform/terraform_configurator.py
```diff
@@ -70,10 +75,26 @@ def __new_aws_instance(self, instance):
...
+        if self.cloud == 'aws':
+            self.__new_aws_key_pair(region)
+
+    def __new_aws_key_pair(self, region):
+        key_name = f'{region}-key'
+        
+        new_key_pair = {
+            'provider': f'aws.{region}',
+            'key_name': key_name,
+            'public_key': f'${{file(\"{self.ssh_key_path}.pub")}}',
```
narmaku, line 94: Q: Do we need to escape the first double quotes?

cloud/terraform/terraform_configurator.py
```diff
@@ -115,3 +116,15 @@ def remove_configuration(self):
 
     def set_aws_profile(self, profile):
         self.aws_profile = profile
+
+    def get_username_by_instance_name(self, name):
```
narmaku, line 120: What if we search for AMI id? They should be unique and we don't need to format them with a char replacement.

narmaku: I suggest renaming to:
```suggestion
    def get_instance_username_by_ami_id(self, ami_id):
```

cloud/terraform/terraform_configurator.py
```diff
@@ -20,14 +20,15 @@ def __init__(self, cloud, ssh_key_path):
             }
 
     def configure_from_resources_json(self, resources_path):
+        self.resources_path = resources_path
```
narmaku, line 23: I think we should add `resources_path` to the constructor.

cloud/terraform/terraform_controller.py
```diff
@@ -1,13 +1,15 @@
 import os
 import json
 from typing import final
```
narmaku, line 3: I think we are not using this library in the code. The same for the next line.

cloud/terraform/terraform_controller.py
```diff
@@ -32,12 +34,17 @@ def get_instances(self):
             for resource in resources:
                 if 'aws_instance' not in resource['address']:
                     continue
+
+                username = self.tf_configurator.get_username_by_instance_name(
```
narmaku, line 38: Yep, maybe it makes sense to use the "ami" key instead.

narmaku (changes requested):
Looking good! Some minor comments.

cloud/terraform/terraform_controller.py
jabia99: That way we cannot have multiple instances from the same ami (we need this if we want to test instance type)

cloud/terraform/terraform_controller.py
jabia99: I don't even know where those came from

cloud/terraform/terraform_controller.py
narmaku: Ooooh, that's true

cloud/terraform/terraform_controller.py
```diff
@@ -23,18 +26,25 @@ def get_instances(self):
...
 
         instances_info = {}
         if self.cloud == 'aws':
             # 'address' corresponds to the tf resource id
             for resource in resources:
+                if 'aws_instance' not in resource['address']:
+                    continue
+
+                username = self.tf_configurator.get_username_by_instance_name(
+                    resource['address'].replace('aws_instance.', '')
```
narmaku, line 39: I would do a split using the dot, and then get the second item of the array.
```suggestion
                    resource['address'].split('.')[1]
```

cloud/terraform/terraform_configurator.py
narmaku: Please dismiss this comments.

cloud/terraform/terraform_configurator.py
jabia99: Yes, if not, everything will get messed up when we dump this to the json

narmaku: So, should we escape the second double quotes?
```suggestion
            'public_key': f'${{file(\"{self.ssh_key_path}.pub\")}}',
```

jabia99: Okay, I tried again and we don't need to scape the double quotes 🤷

narmaku (approved):
Approved!

PR #7, opened 2022-04-05
Improved Terraform modules and suite runner.

narmaku:
Also added test cases migrated from ami-val module.

### DIFF
--- a/test/test_suite_runner.py
+++ b/test/test_suite_runner.py
@@ -6,7 +6,7 @@ from test_suite.suite_runner import SuiteRunner
 class TestSuiteRunner:
     test_cloud_provider = 'aws'
     test_instances = ['test instance 1', 'test instance 2']
-    test_ssh_config = '/path/to/ssh/key.pem'
+    test_ssh_config = '/path/to/ssh_config'
     test_output_filepath = 'test/output/filepath.xml'
 
     @pytest.fixture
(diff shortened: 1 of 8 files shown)


test/test_suite_runner.py
```diff
@@ -5,15 +5,15 @@
 
 class TestSuiteRunner:
     test_cloud_provider = 'aws'
```
jabia99, line 7: Shouldn't we get this from resources.json? I know it's just for debugging purposes, but I think we could implement it now

test_suite/suite_runner.py
```diff
@@ -20,17 +30,23 @@ def compose_testinfra_command(self, output_filepath):
...
         ]
 
+        if self.parallel:
+            command_with_args.append(f'--numprocesses={len(self.instances)}')
+            command_with_args.append(f'--maxprocesses={self.max_processes}')
+
+            # rerun failed tests in case ssh times out or connection is refused by host
+            command_with_args.append('--only-rerun="refused|timeout"')
+            command_with_args.append('--reruns 3')
+            command_with_args.append('--reruns-delay 5')
```
jabia99, line 45: I would set `reruns` and `reruns-delay`as variables, so its easier to parametrize and consistent with `max_processes`,

jabia99 (commented):
LGTM! Just a few comments

jabia99 (approved):
LGTM!

PR #8, opened 2022-04-13
Feature/add cloud image val main script

narmaku:
Added the main script that can be run locally.

### DIFF
--- a/main/cloud_image_validator.py
+++ b/main/cloud_image_validator.py
@@ -7,7 +7,6 @@ from cloud.terraform.terraform_configurator import TerraformConfigurator
 from lib import ssh_lib
 from test_suite.suite_runner import SuiteRunner
 from result.reporter import Reporter
-from pprint import pprint
 
 
 class CloudImageValidator:
@@ -56,11 +55,6 @@ class CloudImageValidator:
         self.infra_controller.create_infra()
         instances = self.infra_controller.get_instances()
 
-        if self.debug:
-            pprint(instances)
-            with open('/tmp/created_instances.json', 'w') as f:
-                json.dump(instances, f)
-
         ssh_lib.generate_instances_ssh_config(instances=instances,
                                               ssh_config_file=self.ssh_config_file,
                                               ssh_key_path=self.ssh_identity_file)
@@ -69,9 +63,9 @@ class CloudImageValidator:
 
     def run_tests_in_all_instances(self, instances):
         time.sleep(5)
-        runner = SuiteRunner(self.infra_configurator.cloud,
-                             instances,
-                             self.ssh_config_file,
+        runner = SuiteRunner(cloud_provider=self.infra_configurator.cloud,
+                             instances=instances,
+                             ssh_config=self.ssh_config_file,
                              parallel=self.parallel,
                              debug=self.debug)
         runner.run_tests(self.output_file)
--- a/main/cloud_image_validator.py
+++ b/main/cloud_image_validator.py
@@ -18,26 +18,13 @@ class CloudImageValidator:
     ssh_config_file = '/tmp/ssh_config'
 
     infra_controller = None
+    infra_configurator = None
 
     def __init__(self, resources_file, output_file, parallel, debug):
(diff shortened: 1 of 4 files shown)


main/cloud_image_validator.py
```diff
@@ -0,0 +1,84 @@
...
+        reporter.generate_html_report(self.output_file.replace('xml', 'html'))
+
+    def cleanup(self):
+        self.infra_controller.destroy_infra()
+
+        os.remove(self.ssh_identity_file)
+        os.remove(self.ssh_pub_key_file)
+
+        if not self.debug:
+            os.remove(self.ssh_config_file)
```
jabia99, line 84: Thinking about logs in the long term, maybe we can move this file to `/tmp/cloud_image_val_logs/` os something like this. Having one place where we store all relevant logs and files. What do you think?

main/cloud_image_validator.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            instances = self.deploy_infrastructure()
+
+            self.run_tests_in_all_instances(instances)
+
+            self.report_test_results()
+        finally:
+            self.cleanup()
+
+    def initialize_infrastructure(self):
```
jabia99, line 40: I think we might have to change other modules, but we have functions that are completely private to the class like this one. Should we use the `__function` convention? What is the standard?

main/cloud_image_validator.py
```diff
@@ -0,0 +1,84 @@
...
+        instances = self.infra_controller.get_instances()
+
+        ssh_lib.generate_instances_ssh_config(instances=instances,
+                                              ssh_config_file=self.ssh_config_file,
+                                              ssh_key_path=self.ssh_identity_file)
+
+        return instances
+
+    def run_tests_in_all_instances(self, instances):
+        time.sleep(5)
```
jabia99, line 65: A comment on why this is needed (the error you were getting) might be useful

test/test_cloud_image_validator.py
```diff
@@ -0,0 +1,133 @@
...
+from result.reporter import Reporter
+from test_suite.suite_runner import SuiteRunner
+
+
+class TestCloudImageValidator:
+    test_resources_file = '/fake/test/resources_file.json'
+    test_output_file = '/fake/test/output_file.xml'
+    test_parallel = True
+    test_debug = True
+    test_instances = ['test instance 1', 'test instance 2']
```
jabia99, line 15: I think I'm being very picky, but it triggers my OCD to see space separated names instead of underscore separation :S

jabia99 (approved):
Looks great! Just a few comments

main/cloud_image_validator.py
narmaku: Not sure right know where is the best place to put these files, since they can be used for debugging and they are not log files _per se_. Maybe we can raise this point when the implement logging capabilities (in the future).

main/cloud_image_validator.py
narmaku: This will prevent several test cases to initially fail due to ssh connection timeout. It's true that maybe here it's not the best place to put the sleep. Let's see if we can move it to another place in the future.

main/cloud_image_validator.py
```diff
@@ -0,0 +1,84 @@
...
+
+class CloudImageValidator:
+    supported_providers = ('aws')
+
+    ssh_identity_file = '/tmp/ssh_key'
+    ssh_pub_key_file = f'{ssh_identity_file}.pub'
+    ssh_config_file = '/tmp/ssh_config'
+
+    infra_controller = None
+    infra_configurator = None
```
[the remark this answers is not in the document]
narmaku: I like to initialize the attributes. It's more convenient when creating unit tests :P

main/cloud_image_validator.py
narmaku: It's up to us to decide. I don't have a strong opinion on that, unless it's clearly a private method.
E.g. we still don't know if we will reuse these functions for the Jenkins CI (being called from different stages).

PR #15, opened 2022-04-22
Migrated some tests from ami-val and os-tests.

narmaku:


test_suite/generic/test_generic.py
```diff
@@ -9,6 +12,57 @@ def test_bash_history_empty(self, host):
                 file_content_length = len(bash_history_file.content_string)
                 assert file_content_length == 0, f'{file_path} must be empty or nonexistent'
 
+    def test_console_is_redirected_to_ttys0(self, host):
+        assert host.file('/proc/cmdline').contains('console=ttyS0'),\
+            'Serial console should be redirected to ttyS0'
+
+    def test_crashkernel_is_enabled_rhel_6(self, host):
```
jabia99, line 19: I just thought maybe we could use pytest tags for tests that are for specific version. I'm not saying we should do it now, just something to think about in the future.

jabia99 (approved):
Looks great!

test_suite/generic/test_generic.py
narmaku: Definitely, it sounds like the way to go. Let's do it in the next migration iteration!

test_suite/cloud/test_aws.py
```diff
@@ -1,14 +1,50 @@
...
+
+
+class TestsNetworkDrivers:
+    def test_correct_network_driver_is_used(self, host):
+        with host.sudo():
+            if not host.package('lshw').is_installed:
+                host.run_test('dnf install lshw -y')
+
+            nic_name = host.check_output('lshw -C network')
+            nic_driver = host.check_output('lshw -C network | grep "driver="')
```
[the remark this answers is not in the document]
narmaku: Well spotted! Actually it's not a path anymore (before when doing testing it was, but then I changed the implementation). So it makes sense to remove the "path" of the arg in next function. I will do it.

PR #19, opened 2022-04-27
Added test_lib and migrated more tests from ami-val.

narmaku:


lib/test_lib.py
```diff
@@ -0,0 +1,17 @@
+def is_rhel_atomic_host(host):
+    with host.sudo():
+        host.file('/etc/redhat-release').contains('Atomic')
```
jabia99, line 3: Just curiosity why no return statement?

test_suite/cloud/test_aws.py
```diff
@@ -53,6 +55,102 @@ def test_nouveau_is_blacklisted(self, host):
         assert host.file(file_to_check).contains('blacklist nouveau'), \
             f'nouveau is not blacklisted in "{file_to_check}"'
 
+    def test_unwanted_packages_are_not_present(self, host):
+        unwanted_pkgs = [
```
jabia99, line 59: I'm guessing some of this packages are also going to be part of a generic test. Or are all of them specific to aws?

lib/test_lib.py
narmaku: Because I forgot :rofl: good catch!

test_suite/cloud/test_aws.py
narmaku: AFAIK, they are specific to AWS. However there might be some of them that are common to other clouds.
We won't know until we implement the same tests for other providers.

lib/test_lib.py
```diff
@@ -0,0 +1,17 @@
+def is_rhel_atomic_host(host):
+    with host.sudo():
+        return host.file('/etc/redhat-release').contains('Atomic')
+
+
+def is_rhel_sap(host):
```
jabia99, line 6: we need a return here to, right?

lib/test_lib.py
```diff
@@ -0,0 +1,17 @@
...
+    __test_keyword_in_repositories(host, 'sap-bundle')
+
+
+def is_rhel_high_availability(host):
```
jabia99, line 10: same as above

narmaku: True

lib/test_lib.py
narmaku: True

PR #20, opened 2022-05-05
Improved suite runner to use default connectionr (paramiko)

narmaku:
Also improved failed test retry filter and added pprint of the generated instances in cloud_image_validator.


jabia99 (approved):
LGTM! with this update cloud-image-val will fly 🚀

narmaku (commented):
Approved.

PR #21, opened 2022-05-06
Improved html reporting

narmaku:
And fixed an issue with relative import of a module.

### DIFF
--- a/cloud-image-val.py
+++ b/cloud-image-val.py
@@ -1,3 +1,5 @@
+import os
+
 from argparse import ArgumentParser, RawTextHelpFormatter
 from main.cloud_image_validator import CloudImageValidator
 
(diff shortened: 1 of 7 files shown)


cloud-image-val.py
```diff
@@ -29,6 +31,8 @@
 if __name__ == '__main__':
     args = parser.parse_args()
 
+    os.environ['PYTHONPATH'] = ':'.join([f'{os.path.dirname(__file__)}', os.environ['PYTHONPATH']])
```
jabia99, line 34: Did this solve the import issue? Maybe it would be nice adding a comment or mention the issue on the commit message

jabia99 (approved):
Looks great, I like the simplicity of pytest-html

PR #23, opened 2022-05-09
migrated two tests from ami-val project

jabia99:
Migrated two basic tests from ami-val

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -208,7 +208,7 @@ class TestsAWS:
         files_to_check = ['ssh_host_ecdsa_key', 'ssh_host_ed25519_key', 'ssh_host_rsa_key']
         for file in files_to_check:
             if host.file(f'/etc/ssh/{file}').exists:
-                assert host.file('/etc/ssh/').mode >= 0o640, \
+                assert host.file(f'/etc/ssh/{file}').mode >= 0o640, \
                     'ssh files permissions are not set correctly'
 
 
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -203,11 +203,13 @@ class TestsAWS:
 
     def test_hostkey_permissions(self, host):
         """
-        bz:2013644
+        BugZilla 2013644
         """
-        ssh_permissions = host.check_output('ls -l /etc/ssh/{ssh_host_ecdsa_key,ssh_host_ed25519_key,ssh_host_rsa_key}')
-        assert '-rw-------. 1 root root' not in ssh_permissions, \
-            'ssh files permissions are not set correctly'
+        files_to_check = ['ssh_host_ecdsa_key', 'ssh_host_ed25519_key', 'ssh_host_rsa_key']
+        for file in files_to_check:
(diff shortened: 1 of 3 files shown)


test_suite/generic/test_generic.py
```diff
@@ -135,6 +135,14 @@ class TestsNetworking:
     def test_dns_resolving_works(self, host):
         host.run_test('ping -c 5 google-public-dns-a.google.com')
 
+    def test_ipv_localhost(self, host):
+        expected_hosts = ['127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4',
+                          '::1         localhost localhost.localdomain localhost6 localhost6.localdomain6']
+        with host.sudo():
+            for expected_host in expected_hosts:
+                assert host.file('/proc/cmdline').contains(expected_host), \
+                    "/etc/hosts does not contain ipv4 or ipv6 localhost"
```
narmaku, line 144:
```suggestion
                    '/etc/hosts does not contain ipv4 or ipv6 localhost'
```

test_suite/generic/test_generic.py
```diff
@@ -135,6 +135,14 @@ class TestsNetworking:
     def test_dns_resolving_works(self, host):
         host.run_test('ping -c 5 google-public-dns-a.google.com')
 
+    def test_ipv_localhost(self, host):
+        expected_hosts = ['127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4',
+                          '::1         localhost localhost.localdomain localhost6 localhost6.localdomain6']
+        with host.sudo():
+            for expected_host in expected_hosts:
+                assert host.file('/etc/hosts').contains(expected_host), \
```
narmaku, line 143: I think we are reading the wrong file here :sweat_smile:

narmaku (changes requested):
Nice. Some minor comments!

test_suite/generic/test_generic.py
jabia99: why?

test_suite/cloud/test_aws.py
```diff
@@ -201,6 +201,16 @@ def test_max_cstate_is_configured_in_cmdline(self, host):
...
                         f'{line} must not be specified in AMIs that are not SAP'
 
+    def test_hostkey_permissions(self, host):
+        """
+        BugZilla 2013644
+        """
+        files_to_check = ['ssh_host_ecdsa_key', 'ssh_host_ed25519_key', 'ssh_host_rsa_key']
+        for file in files_to_check:
+            if host.file(f'/etc/ssh/{file}').exists:
+                assert host.file('/etc/ssh/').mode >= 0o640, \
```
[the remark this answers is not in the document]
narmaku: Question: Does the `0o640` mode also check the group and the owner (I think root/root was the unwanted one)?

If not, we will have to add that check too (see File.user and File.group).

narmaku (changes requested):

test_suite/cloud/test_aws.py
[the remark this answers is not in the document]
jabia99: oh right, I think it does get checked but it doesn't make sense to do it that way

narmaku (approved):
LGTM!

PR #24, opened 2022-05-11
Added two more tests and other improvements.

narmaku:
Added global fixture to get instance data from new instances.json, generated when deploying infrastructure.
Added error message to html report.

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -1,3 +1,4 @@
+import json
 import re
 import pytest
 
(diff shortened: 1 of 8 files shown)


test_suite/generic/test_generic.py
```diff
@@ -114,6 +114,25 @@ def test_cert_product_version_is_correct(self, host):
...
+        with host.sudo():
+            if host.package('systemd').is_installed:
+                assert '/lib/systemd/system/multi-user.target' in \
+                       host.check_output('readlink -f /etc/systemd/system/default.target'), \
+                    'Unexpected systemd default target'
+            else:
+                assert 'id:3:initdefault' in host.check_output("grep '^id:' /etc/inittab"), \
+                    'Unexpected default inittab "id"'
+
+                if 'el5' in kernel_release:
```
jabia99, line 132: I thought we said we were not going to support rhel 6 or lower. I doesn´t hurt to have this though

test_suite/cloud/test_aws.py
```diff
@@ -211,6 +212,44 @@ def test_hostkey_permissions(self, host):
                 assert host.file(f'/etc/ssh/{file}').mode >= 0o640, \
                     'ssh files permissions are not set correctly'
 
+    def test_aws_instance_identity(self, host, instance_data):
```
jabia99, line 215: Just curious, why don´t we include ami-val docstrings? I think the save some time when trying to understand the test

jabia99 (approved):
looks good

jabia99 (approved):
Looks good

test_suite/cloud/test_aws.py
narmaku: Good point. It makes even more sense now that we have the description of the HTML report directly obtained from doctstrings. Let's do it for all the migrated tests! I will review the currently migrated ones and update them accordingly (in a new commit, to ease the review).

test_suite/generic/test_generic.py
narmaku: True. I replicated the same logic that is in the original test, but we can ask Frank if this is still really needed or not (I would suppose it's not). In the case RHEL 6.0 is not going to be supported anymore, we will have to update some of the tests already migrated (not a big deal IMO).

test_suite/cloud/test_aws.py
jabia99: 🚀🚀🚀

PR #26, opened 2022-05-11
Migrated two tests.

narmaku:
Also added a fixture for AWS instances data to be obtained from their web service.


jabia99 (approved):
LGTM! Thanks for changing `TestAWSNetworking` class name, it's much more clear now

PR #27, opened 2022-05-12
Improved HTML report by adding "Image Reference" for each test case.

narmaku:
Also migrated one more test case from ami-val.


jabia99 (approved):
LGTM!

PR #28, opened 2022-05-13
migrate check_rhel_version

jabia99:


test_suite/generic/test_generic.py
```diff
@@ -151,6 +152,27 @@ def test_inittab_and_systemd(self, host):
...
+        release_file = 'redhat-release'
+        if host.system_info.distribution == 'fedora':
+            release_file = 'fedora-release'
+
+        with host.sudo():
+            package_release_version = host.check_output("sudo rpm -q --qf '%{VERSION}' --whatprovides " + release_file)
+
+        assert product_version == package_release_version, \
+            f'product version ({product_version}) does not match package release version'
+        assert product_version.replace('.', '-') in cloud_image_name, 'product version in not in image name'
```
narmaku, line 174:
```suggestion
        assert str(product_version).replace('.', '-') in cloud_image_name, 'product version is not in image name'
```

test_suite/generic/test_generic.py
```diff
@@ -151,6 +152,27 @@ def test_inittab_and_systemd(self, host):
...
+        product_version = host.system_info.release
+
+        release_file = 'redhat-release'
+        if host.system_info.distribution == 'fedora':
+            release_file = 'fedora-release'
+
+        with host.sudo():
+            package_release_version = host.check_output("sudo rpm -q --qf '%{VERSION}' --whatprovides " + release_file)
+
+        assert product_version == package_release_version, \
```
narmaku, line 172:
```suggestion
        assert product_version == float(package_release_version), \
```

test_suite/generic/test_generic.py
```diff
@@ -151,6 +152,27 @@ def test_inittab_and_systemd(self, host):
...
+
+        cloud_image_name = instance_data['name']
+        product_version = host.system_info.release
+
+        release_file = 'redhat-release'
+        if host.system_info.distribution == 'fedora':
+            release_file = 'fedora-release'
+
+        with host.sudo():
+            package_release_version = host.check_output("sudo rpm -q --qf '%{VERSION}' --whatprovides " + release_file)
```
narmaku, line 170: No need to call sudo with sudo.

```suggestion
            package_release_version = host.check_output("rpm -q --qf '%{VERSION}' --whatprovides " + release_file)
```

narmaku, line 170: I think there's a way to check package version with a testinfra module.

See: https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.package.Package.version

test_suite/generic/test_generic.py
```diff
@@ -151,6 +152,27 @@ def test_inittab_and_systemd(self, host):
                     assert 'si::sysinit:/etc/rc.d/rc.sysinit' in host.check_output("grep '^si:' /etc/inittab"), \
                         'Unexpected default inittab "id"'
 
+    def test_release_version(self, host, instance_data):
+        '''
```
narmaku, line 156: Replace single quotes by double quotes for multi-line strings and docstrings. I think flake8 also highlights this.

test_suite/generic/test_generic.py
```diff
@@ -1,5 +1,6 @@
 import pytest
-
+from pytest_mock import package_mocker
```
narmaku, line 2: What is this ninja artifact?

narmaku (changes requested):

test_suite/generic/test_generic.py
jabia99: upsss

test_suite/generic/test_generic.py
jabia99: Can't be used here as we are using --whatprovides option it's not available on test_infra

test_suite/generic/test_generic.py
jabia99: product version is already a string, I tested it and passed on all images

test_suite/generic/test_generic.py
jabia99: we are comparing string to string already. Would it be better comparing floats?

narmaku: I see, and it's an "equals" comparison so it doesn't matter. Please dismiss this comment!

test_suite/generic/test_generic.py
narmaku: I see!

test_suite/generic/test_generic.py
narmaku: I see, it makes sense then to keep it like it was! Sorry about it.

narmaku (approved):
LGTM!

PR #29, opened 2022-05-13
add test_sshd

jabia99:
ami-val tests that the instance uses systemctl. As cloud-image-val does not support rhel-6 or lower, this is not necessary.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -1,6 +1,5 @@
 import pytest
 from lib import test_lib
-import re
 
 
 class TestsGeneric:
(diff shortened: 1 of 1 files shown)


test_suite/generic/test_generic.py
```diff
@@ -173,6 +173,18 @@ def test_release_version(self, host, instance_data):
...
+    def test_sshd(self, host):
+        """
+        sshd service shoud be on, password authentication shoud be disabled
+        """
+        with host.sudo():
+            sshd = host.service('sshd')
+            assert sshd.is_running, 'ssh.service is not active'
+
+            if not host.file('/etc/ssh/sshd_config').contains('PasswordAuthentication no') and \
+               not host.file('/etc/ssh/sshd_config').contains('#PasswordAuthentication yes'):
```
narmaku, line 185: We could be pickier in this test.

Check what is being done here:
https://github.com/osbuild/cloud-image-val/blob/4944549b6a824d32c2e4e992f03b7e096df2422d/test_suite/cloud/test_aws.py#L220

narmaku (changes requested):

test_suite/generic/test_generic.py
```diff
@@ -181,8 +182,12 @@ def test_sshd(self, host):
             sshd = host.service('sshd')
             assert sshd.is_running, 'ssh.service is not active'
 
-            if not host.file('/etc/ssh/sshd_config').contains('PasswordAuthentication no') and \
-               not host.file('/etc/ssh/sshd_config').contains('#PasswordAuthentication yes'):
+            sshd_config_string = host.file('/etc/ssh/sshd_config').content_string
+            pwdAuthNo = 'PasswordAuthentication no'
+            pwdAuthYes = 'PasswordAuthentication yes'
+
+            if not re.match(pwdAuthNo, sshd_config_string) and \
```
narmaku, line 189: Actually, how the code is done now, then we will have a false positive in the case the following line is present:
`PasswordAuthentication yes`

I think we can replace the `if` by an `assert` to follow what is being done in the other tests. In that case, the condition statements become "positive" and you avoid negations.

```
assert (re.match(pwdAuthNo, sshd_config_string) and 
    not re.match(pwdAuthYes, sshd_config_string)) or \
    re.match(f'#[ ]?{pwdAuthYes}|#[ ]+{pwdAuthYes}', sshd_config_string), \
    'Password authentication should be disabled'
```

narmaku (changes requested):

test_suite/generic/test_generic.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 from lib import test_lib
+import re
```
narmaku, line 3:
```suggestion
```

narmaku (approved):
LGTM. We just need to remove the `import re` line.

narmaku (approved):

jabia99:
Thanks for helping in this incredibly short but incredibly time consuming test!

PR #37, opened 2022-05-24
Added markers option to run tests matching pytest markers expression.

narmaku:
Added rhel_only fixture and updated tests cases that are only applicable to RHEL images.
Marked test cases as "pub", which are only applicable to published cloud images.
Improved HTML report with new summary element to know which markers have been used.
Improved HTML report column "Test Case" so that it displays a shorter name of the test executed.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+markers =
+    pub: Tests that have to be run on images that are already published into cloud providers' marketplaces.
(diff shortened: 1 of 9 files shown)


pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    pub: Tests that have to be run on images that are already published into cloud providers' marketplaces.
```
jabia99, line 3: Apostrophe typo

jabia99 (commented):
Looks great! Does the fixture work as expected?

pytest.ini
narmaku: Ad discussed via Slack, this is also valid grammatically speaking.

narmaku:
> Looks great! Does the fixture work as expected?

@jabia99 , yes, but remember this is not a fixture, but a marker :+1:

PR #40, opened 2022-05-26
divide test_release_version in two different tests

jabia99:
test_release_version had two assertions. One of them was related to the
image name, which only makes sense in published images. Divide test so
we can use the 'pub' mark.

Also add `rhel_only` fixture to `test_wheel_group_not_set_to_default_use`

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -309,7 +309,7 @@ class TestsCloudInit:
         assert host.file('/etc/cloud/cloud.cfg').contains('- growpart'), \
             'growpart must be present in cloud_init_modules'
 
-    def test_wheel_group_not_set_to_default_user(self, host):
+    def test_wheel_group_not_set_to_default_user(self, host, rhel_only):
         """
         Make sure there is no wheel in default_user's group in "/etc/cloud/cloud.cfg".
         BugZilla 1549638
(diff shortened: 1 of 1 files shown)


jabia99:
I may have missed this on a previous PR. But why do we have rhel-only fixture on this test @narmaku ?

narmaku:
> I may have missed this on a previous PR. But why do we have rhel-only fixture on this test @narmaku ?

@jabia99 , because in the original the test case we were checking specifically for a Red Hat specific file.

Now that the test has been decoupled (nice!), I think we can remove `rhel_only` from `test_release_version_in_image_name`. I think we are publishing all the images with their version in the AMI name, regardless their distro.

test_suite/generic/test_generic.py
```diff
@@ -172,6 +171,18 @@ def test_release_version(self, host, instance_data, rhel_only):
 
         assert product_version == package_release_version, \
             f'product version ({product_version}) does not match package release version'
+
+    @pytest.mark.pub
+    def test_release_version_in_image_name(self, host, instance_data, rhel_only):
```
narmaku, line 176: Not needed here now that the original test has been decoupled.

```suggestion
    def test_release_version_in_image_name(self, host, instance_data):
```

narmaku (commented):
LGTM! Only one thing: Maybe we should remove the `rhel_only` fixture from the second test case.

jabia99:
Are you sure it is a rhel only file? we are actually checking if the OS is fedora inside the test

test_suite/generic/test_generic.py
jabia99: Great!

PR #45, opened 2022-05-30
add parameter to stop cleanup process

ghost:
when debugging locally, it's sometimes useful to connect to the
instances via ssh. Stoping the cleanup process makes this possible.

### DIFF
--- a/main/cloud_image_validator.py
+++ b/main/cloud_image_validator.py
@@ -25,7 +25,8 @@ class CloudImageValidator:
                  test_filter=None,
                  include_markers=None,
                  parallel=False,
-                 debug=False):
+                 debug=False,
+                 stop_cleanup=False):
         self.resources_file = resources_file
         self.output_file = output_file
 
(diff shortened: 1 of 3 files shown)


main/cloud_image_validator.py
```diff
@@ -45,6 +47,9 @@ def main(self):
             console_lib.print_divider('Running tests')
             self.run_tests_in_all_instances(instances)
         finally:
+            if self.stop_cleanup:
+                console_lib.print_divider('Cleanup stopped, press any key to continue')
```
narmaku, line 51: Maybe this divider is redundant IMO. We can just leave the input message take care of letting the user know that CIV is waiting for user interaction to continue with the cleanup (see suggestion below).

main/cloud_image_validator.py
```diff
@@ -45,6 +47,9 @@ def main(self):
...
+                input('Proceed with cleanup?')
```
narmaku, line 52:
```suggestion
                input('Cleanup stopped. Press any key to proceed with the cleanup')
```

PR #52, opened 2022-06-02
Changed logic of randomizing resource names.

narmaku:
Minor improvements when stopping cleanup: Now we print the ssh commands to use for each image name.

### DIFF
--- a/cloud/terraform/aws_config_builder.py
+++ b/cloud/terraform/aws_config_builder.py
@@ -35,13 +35,15 @@ class AWSConfigBuilder(BaseConfigBuilder):
         self.resources_tf['resource']['aws_instance'] = {}
 
         for instance in self.resources_dict['instances']:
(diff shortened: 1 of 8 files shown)


cloud/terraform/aws_config_builder.py
```diff
@@ -35,13 +35,15 @@ def build_resources(self):
...
+            self.__new_aws_key_pair(instance)
             self.__new_aws_instance(instance)
 
         return self.resources_tf
 
-    def __new_aws_key_pair(self, region):
-        key_name = f'{region}-key{self.random_str}'
+    def __new_aws_key_pair(self, instance):
+        region = instance['region']
+        key_name = self.build_resource_name([instance['region'], 'key'])
```
jabia99, line 45: just being picky. Maybe `build` is not the best word for this function(?) What do you think of `create_resource_name`

jabia99 (commented):
Looks great!

PR #53, opened 2022-06-02
Added ssh-keyscan wait for all hosts in multi-threading.

narmaku:


lib/ssh_lib.py
```diff
@@ -26,3 +27,23 @@ def generate_instances_ssh_config(ssh_key_path, ssh_config_file, instances):
...
+                            If timeout is reached, an exception will be thrown.
+    :return: None
+    """
+    count = 0
+    while count < timeout_seconds:
+        if (os.system(f'ssh-keyscan "{host}" > /dev/null 2>&1') >> 8) == 0:
+            print(f'{host} SSH is up! ({count} seconds)')
+            break
+        else:
+            time.sleep(1)
```
jabia99, line 46: I might be being a bit picky here, we can approve as it is. But waiting this way we are going to wait more that `timeout_seconds`, as ssh-keyscan takes same time, even if it's small.

jabia99 (approved):

jabia99 (commented):
Suggestion on how to measure the time elapsed

jabia99:
The commit history is a bit messed up. I would drop commit d98d752e63d7abb9408c4c04080e45a9ced3d49f 

Other than that, looks great!

narmaku:
> The commit history is a bit messed up. I would drop commit [d98d752](https://github.com/osbuild/cloud-image-val/commit/d98d752e63d7abb9408c4c04080e45a9ced3d49f)
> 
> Other than that, looks great!

Yeah it was my bad, I made a mistake when re-basing locally then I had to add it as a commit.

PR #54, opened 2022-06-06
add tag to aws resources so they can be removed by SCC

jabia99:


cloud/terraform/aws_config_builder.py
```diff
@@ -69,7 +70,7 @@ def __new_aws_instance(self, instance):
             'ami': instance['ami'],
             'provider': f'aws.{instance["region"]}',
             'key_name': instance['aws_key_pair'],
-            'tags': {'name': name_tag},
+            'tags': {'name': name_tag, 'gitlab-ci-test': 'true'},
```
narmaku, line 73: This tag will be shared among all cloud providers. Right?

If that will be the case, I suggest adding an attribute to the BaseConfigBuilder class and use it here as a key.

E.g.:
```suggestion
            'tags': {'name': name_tag, self.ci_tag_key: 'true'},
```

jabia99:
> Can we also add tags to Azure resources?

Scheduled cloud cleaner is much more complicated for now. I say we leave this for when we have azure running in some automated CI, as it is not high priority (and we might change how  we deal with azure cleanup)

PR #55, opened 2022-06-07
Fix markers logic

narmaku:
Fixed logic to detect which markers are correct and covered the case when the test has no markers.
Also fixed Azure VM's public dns in Terraform Controller at the moment of getting instances data (get_instances()).


jabia99 (approved):
Other than the unit test failure. LGTM!

PR #58, opened 2022-06-08
Migrated more tests

narmaku:


test_suite/cloud/test_aws.py
```diff
@@ -397,6 +397,54 @@ def test_yum_plugins(self, host):
...
+                f'Expected configuration was not found in "{file_to_check}"'
+
+    @pytest.mark.run_on(['rhel8.5', 'rhel8.6', 'rhel9.0'])
+    def test_dracut_conf_xen(self, host):
+        """
+        BugZilla 1849082
+        Add ' xen-netfront xen-blkfront ' to '/etc/dracut.conf.d/xen.conf' in x86 AMIs prior RHEL-8.4.
+        Using image builder from RHEL-8.5, add ' nvme xen-blkfront ' to '/usr/lib/dracut/dracut.conf.d/ec2.conf'.
+        This is not required in arm AMIs.
+        """
```
jabia99, line 423: I would not remove the ticket reference `#COMPOSER-1096` from the test. I think every context we can provide has value. Maybe we can include it on the docstring

jabia99 (approved):
LGTM!

test_suite/cloud/test_aws.py
narmaku: Good point! I forgot to add it here.

jabia99 (approved):
Looks great!

The subscription manager test is huge! Did you test it on all sample images?

narmaku:
> Looks great!
> 
> The subscription manager test is huge! Did you test it on all sample images?

Thanks! Yes I did, but the tests will fail in all AMIs except some of them that are ready to be published. E.g. the one that Frank shared with us via chat. In that AMI, there's a timeout while checking the subscription status, which is always unknown. I think it's something related to the environment and not the test itself. So we will follow-up on this with Frank directly.

PR #59, opened 2022-06-13
Migrated some more test cases from ami-val.

narmaku:
Also fixed a false positive in cpu flags test case.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -1,4 +1,5 @@
 import pytest
+
 from lib import test_lib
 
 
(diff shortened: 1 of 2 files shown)


test_suite/generic/test_generic.py
```diff
@@ -14,6 +14,14 @@ def test_bash_history_is_empty(self, host):
                 file_content_length = len(bash_history_file.content_string)
                 assert file_content_length == 0, f'{file_path} must be empty or nonexistent'
 
+    @pytest.mark.run_on(['rhel'])
```
jabia99, line 17: I would add fedora to the test. And also investigate centos

jabia99 (commented):
Just a small comment

test_suite/generic/test_generic.py
narmaku: I added a TODO for the future, since it's not in the priorities now.

jabia99 (approved):
LGTM!

PR #60, opened 2022-06-14
Improved azure config builder

narmaku:
This update brings the possibility of deploying Azure VMs directly from a vhd blob URI.
The Terraform Controller will create an image based on a given vhd, and then create the VM based on that image.

### DIFF
--- a/cloud/terraform/azure_config_builder.py
+++ b/cloud/terraform/azure_config_builder.py
@@ -35,6 +35,8 @@ class AzureConfigBuilder(BaseConfigBuilder):
         self.resources_tf['resource']['azurerm_virtual_network'] = {}
         self.resources_tf['resource']['azurerm_subnet'] = {}
 
+        # VM specific resources
+        self.resources_tf['resource']['azurerm_image'] = {}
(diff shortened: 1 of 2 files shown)


cloud/terraform/azure_config_builder.py
```diff
@@ -46,12 +48,35 @@ def build_resources(self):
...
+        }
+
+        new_subnet = {
+            'name': name,
+            'location': instance['location'],
+            'resource_group_name': self.resource_group,
+            'os_disk': os_disk,
+        }
+
+        self.resources_tf['resource']['azurerm_image'][name] = new_subnet
```
jabia99, line 78: I think variable names are a bit confusing. You are assigning `new_subnet` to an `azurerm_image`. Is this correct?

jabia99 (commented):
Just a small comment

cloud/terraform/azure_config_builder.py
narmaku: LOL - copy&paste effect! I will change it. Well spotted!

jabia99 (approved):
LGTM!

PR #62, opened 2022-06-17
Added support for Google Cloud.

narmaku:


cloud/terraform/terraform_controller.py
```diff
@@ -110,6 +109,28 @@ def get_instances_azure(self, resources):
...
+            if resource['type'] != 'google_compute_instance':
+                continue
+
+            public_ip = resource['values']['network_interface'][0]['access_config'][0]['nat_ip']
+
+            instances_info[resource['address']] = {
+                'name': resource['name'],
+                'instance_id': resource['values']['id'],
+                'public_ip': public_ip,
+                'public_dns': public_ip,
```
jabia99, line 126: No DNS for google images? interesting

cloud/terraform/gcloud_config_builder.py
```diff
@@ -0,0 +1,135 @@
...
+        new_vpc = {
+            'name': network_name,
+            'auto_create_subnetworks': True
+        }
+
+        self.resources_tf['resource']['google_compute_network'][network_name] = new_vpc
+
+    def __new_gcloud_instance(self, instance):
+        if not instance['instance_type']:
+            instance['instance_type'] = 'e2-micro'
```
jabia99, line 87: Are we sure this instance type is enough for our testing? Maybe we should add a TODO so we don't forget to verify it

cloud/terraform/gcloud_config_builder.py
```diff
@@ -0,0 +1,135 @@
...
+            'tags': [self.ssh_enabled_tag, self.ci_tag_key],
+            'depends_on': [
+                'google_compute_network.{}'.format(instance['google_compute_network'])
+            ]
+        }
+
+        self.resources_tf['resource']['google_compute_instance'][name] = new_instance
+
+    def __get_gcloud_resource_self_link(self, resource_name):
+        pass
```
jabia99, line 135: this is still WIP, right?

jabia99 (commented):
Just a few questions

cloud/terraform/gcloud_config_builder.py
```diff
@@ -0,0 +1,135 @@
...
+                'image': instance['image']
+            }
+        }
+
+        network_interface = {
+            'network': instance['google_compute_network'],
+            'access_config': {}
+        }
+
+        if 'username' in instance:
```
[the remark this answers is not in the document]
narmaku: For Azure and GCP it is not needed to specify a username (unlike AWS, which the user name to be used is the one already "embedded" into the AMI). So, in AZ and GCP you can use whatever username you like, but in AWS it has to be yes or yes the one defined inside the AMI.

In Azure we require "username" field to be present in the resources.json but we shouldn't (this is an improvement we have to do). So, in GCP I directly made it optional, just like it should be.

cloud/terraform/gcloud_config_builder.py
narmaku: No, not sure really, but we can also specify the instance type in the resources.json. The same for AZ and AWS. so far, this default instance type has worked well, although only 2 images have been tested. I think we will discover if this is an issue in the future.

jabia99: Cool, let's keep this in mind then

cloud/terraform/terraform_controller.py
jabia99: I don't even know why we'd want a DNS. I was just curious

PR #75, opened 2022-06-24
Migrate ami val test SAP High Availability specific test.

narmaku:


jabia99 (approved):
Let's create a low priority ticket to divide this sh test into smaller python test. But looks great!

PR #76, opened 2022-06-24
Migrated reboot-related tests from ami-val.

narmaku:


test_suite/generic/test_generic.py
```diff
@@ -396,6 +399,7 @@ def test_wheel_group_not_set_to_default_user(self, host):
 
 
 @pytest.mark.pub
+@pytest.mark.order(3)
```
jabia99, line 402: Do we know how the pytest-order would work with parallel execution of tests? (Multiple ssh connection wo the same instance)

jabia99 (approved):
LGTM

test_suite/generic/test_generic.py
narmaku: It will always respect the ordering, even though there are multiple threads.

PR #78, opened 2022-07-01
add ci tag to every azure resource

jabia99:
this way, cloud cleaner can remove the leftover resources in case the is
any


narmaku (approved):
LGTM!

PR #84, opened 2022-07-13
Tagged test to run on rhel 7.9, 8.5, 8.6, fedora 36 and centos.

narmaku:


test_suite/generic/test_generic.py
```diff
@@ -440,7 +440,7 @@ def test_ipv_localhost(self, host):
                 assert host.file('/etc/hosts').contains(expected_host), \
                     '/etc/hosts does not contain ipv4 or ipv6 localhost'
 
-    @pytest.mark.run_on(['all'])
+    @pytest.mark.run_on(['rhel7.9', 'rhel8.5', 'rhel8.6', 'fedora35', 'centos'])
```
atodorov, line 443: According to https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/585856447 it works on 8.7, 9.0, 9.1 as well. Do you have an exclude decorator instead ?

narmaku: Good point. We don't. I think we should implement an exclude decorator as I see this situation coming again in the future.
I'll create a ticket in CLOUDX board for this. For now I will add 8.7, 9.0 and 9.1 to the list.

narmaku: Oh, it's true that we can just add "rhel" and it will apply to all RHEL versions. What do you think about it?

PR #85, opened 2022-07-14
Migrated some os-tests cases.

narmaku:
Also added a higher wait for the reboot test cases, improved instances data with the corresponding cloud provider (so that we can exclude them if needed, in some t est cases), and improved regex to detected which failing test should be rerun.


test_suite/cloud/test_aws.py
```diff
@@ -73,6 +73,55 @@ def test_ami_name(self, host, instance_data):
             assert re.match(fedora_ami_name_format, ami_name), \
                 'Unexpected AMI name for Fedora image'
 
+    @pytest.mark.pub
+    @pytest.mark.run_on(['all'])
+    def test_release_version_in_ami_name(self, host, instance_data):
```
jabia99, line 78: Out of curiosity. Why this test doesn't apply to azure?

data/azure/66-azure-storage.rules
```diff
@@ -0,0 +1,28 @@
+ACTION=="add|change", SUBSYSTEM=="block", ENV{ID_VENDOR}=="Msft", ENV{ID_MODEL}=="Virtual_Disk", GOTO="azure_disk"
```
jabia99, line 1: Do we know if this files are going to get updated? Is this going to imply maintenance?

jabia99 (approved):
Loogs good!

PR #86, opened 2022-07-19
added Dockerfile

jabia99:
CIV needs to be containeraized to avoid dealing with dependency problems
constantly. Added a dockerfile with the first draft of the container


jabia99:
We found a bug when trying to use env varibles. related PR: https://github.com/osbuild/cloud-image-val/pull/87

PR #87, opened 2022-07-19
remove profile from aws command

jabia99:
from now on CIV will use env varibles to authenticate into aws. Removing
the profile lets us use containers.


narmaku (approved):
LGTM!!

PR #92, opened 2022-08-01
update first_boot max time

jabia99:
this test in aws has been failing due to images taking too much time to
boot (~55s instead of the previous max, 50s). raising limit to 60s


narmaku (approved):
LGTM!

PR #95, opened 2022-08-08
Moved test case to the top in the test class, so that it is run first.

narmaku:
This solves the issues experienced with Azure VMs.


jabia99 (approved):
LGTM!

PR #96, opened 2022-08-08
Added new AMIs for AWS, and change the RHEL 7.9 vhd image for Azure.

narmaku:


jabia99 (commented):
Could you change the resource file that we run CIV on our CI? this way we could see how the new AMIs perform ([gitlab-ci line](https://github.com/osbuild/cloud-image-val/blob/2d9871ab669afd46c1b3f8ad8b5b6146a2251b82/.gitlab-ci.yml#L43))

PR #97, opened 2022-08-09
Suppressed output of dnf command when installing lshw utility.

narmaku:
We believe this can avoid overloading stdout and resulting in unexpected -1
return code (testinfra module issue).


jabia99:
Does this fix the error related to -1 exit status?

narmaku:
> Does this fix the error related to -1 exit status?

We are not sure, we hope this can mitigate the issue. We don't know the exact reason why testinfra's CommandResult stores -1 as return code (exit_status) for the command, when it was not even failing. We hope that suppressing the output can solve this, since I also saw this -1 issue when using grep !

PR #98, opened 2022-08-09
Migrated test_check_waagent_log from os-tests.

narmaku:
Also added a test_util function to filter log files. It will be useful
for similar test cases (in os-tests there was also an auxiliary function
for that).


jabia99 (approved):
LGTM!

jabia99:
Did the test run in CI?

jabia99:
Could you build the container with this code so it runs on our CI (I hope this is the last time you have to do it, automation coming!)

PR #99, opened 2022-08-10
Added socket.timeout string to the test retry regex.

narmaku:


jabia99 (commented):
Fix the unit pls, other than that looks okay!

PR #100, opened 2022-08-11
disable test_host_key_permissions on rhel 8.4

jabia99:
rhel 8.4 has an old version of cloud-init, which create hostkey files
with wrong permissions. See COMPOSER-1654 for reference.

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -321,7 +321,7 @@ class TestsAWS:
                     assert not host.file('/proc/cmdline').contains(line), \
                         f'{line} must not be specified in AMIs that are not SAP'
 
-    @pytest.mark.run_on(['rhel', 'centos', 'fedora36'])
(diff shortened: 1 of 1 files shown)


test_suite/cloud/test_aws.py
```diff
@@ -321,7 +321,7 @@ def test_max_cstate_is_configured_in_cmdline(self, host):
                     assert not host.file('/proc/cmdline').contains(line), \
                         f'{line} must not be specified in AMIs that are not SAP'
 
-    @pytest.mark.run_on(['rhel', 'centos', 'fedora36'])
+    @pytest.mark.run_on(['rhel7.9', 'rhel8.4', 'rhel8.5', 'rhel9.0', 'centos', 'fedora36'])
```
narmaku, line 324: Shouldn't we remove ` 'rhel8.4'` from the list?

jabia99: I'm going to blame this error on the heat... Thanks for noticing!

narmaku (approved):
LGTM!

PR #101, opened 2022-08-16
Removed test_udev_kernel

narmaku:


thozza (commented):
I'm sorry, but I think that the whole test case should be thrown away. Checking if `80-net-name-slot.rules` exists on RHEL-8.4 and before just verifies that the file (symlink to `/dev/null`) exists, however the existence of the file has no effect on the system whatsoever. There is no such rule shipped on RHEL-8 so masking it has no effect at all.

The existence of the symlink on RHEL images built using kickstarts can be considered a bug, but without any real impact or effect. Checking that it exists does not add any value.

thozza (approved):
LGTM 👍

PR #103, opened 2022-08-18
Add exclude_on marker

pebringer:


test_suite/conftest.py
```diff
@@ -33,6 +35,20 @@ def check_markers(host, request):
     host_version = host.system_info.release
     distro_version = f'{host_distro}{host_version}'
 
+    # Skip the test if the distro is explicitely excluded
```
narmaku, line 38:
```suggestion
    # Skip the test if the distro is explicitly excluded
```

Typo here.

test_suite/conftest.py
```diff
@@ -33,6 +35,20 @@ def check_markers(host, request):
...
+    if exclude_on_marker:
+        exclude_on_marker_list = exclude_on_marker.args[0]
+        if host_distro in exclude_on_marker_list \
+                or distro_version in exclude_on_marker_list:
+            pytest.skip(f"This test doesn't apply to {distro_version}")
```
narmaku, line 43:
```suggestion
            pytest.skip(f'This test doesn't apply to {distro_version}')
```
We adopted the convention of using single quotes whenever possible (although linters do not care about it).

test_suite/conftest.py
```diff
@@ -33,6 +35,20 @@ def check_markers(host, request):
...
+
+    # If there is no run_on_marker and distro is not excluded execute supported
+    # tests
+    if not run_on_marker:
+        if distro_version in supported_distros_and_versions:
```
narmaku, line 48: I think this `if` is similar to the one in line 61. Do you think we can move the validation here instead?

test_suite/conftest.py
pebringer: Oh I didn't realize that as this part of the code contains the double quotes  on multiple occasions, will keep that in mind!

test_suite/conftest.py
pebringer: I don't think so or maybe I just don't see it. This checks executes the test if there is no run_on marker, the distro is supported and it wasn't filtered out by exclude. Following lines then process the run_on marker as the assumption here is that either run_on, exclude_on or both may be specified.

test_suite/conftest.py
narmaku: Oh, it's true that in this case the double quotes do make sense! Since the "doesn't" apostrophe is there.

test_suite/conftest.py
narmaku: Yeah I was thinking about combining both ifs since the result is similar, but that may add more nesting or complexity.

test_suite/conftest.py
```diff
@@ -33,6 +35,20 @@ def check_markers(host, request):
     host_version = host.system_info.release
     distro_version = f'{host_distro}{host_version}'
 
+    # Skip the test if the distro is explicitly excluded
+    if exclude_on_marker:
+        exclude_on_marker_list = exclude_on_marker.args[0]
+        if host_distro in exclude_on_marker_list \
+                or distro_version in exclude_on_marker_list:
+            pytest.skip(f'This test doesn't apply to {distro_version}')
```
narmaku, line 43:
```suggestion
            pytest.skip(f"This test doesn't apply to {distro_version}")
```
Sorry my mistake, it's true that in this case the double quotes do make sense. I missed the apostrophe inside :sweat:

pebringer: Luckily you noticed! I haven't even try to challenge it since it made sense to me.. sneaky apostrophe!

narmaku (approved):
LGTM!

PR #106, opened 2022-08-18
Added build and push of the updated container.

narmaku:
This applies to jobs for active PRs and main (when a merge occurred).
When there is a merge, container with the 'latest' tag will be updated.
In all cases, the branch-specific container will be built/updated.


jabia99 (approved):
LGTM!

PR #109, opened 2022-08-19
Migrated test_check_nm_cloud_setup test from os-tests.

narmaku:
Also removed the ami-val test equivalent previously migrated,
that was only being run on AWS instances. The test is now generic.


jabia99 (commented):
Seems like the test is failing on rhel 8.5 Could you take a look at it?

jabia99 (approved):
Just solve the conflicts and should be good to go

PR #111, opened 2022-08-23
WIP

narmaku:


narmaku:
> @narmaku what is it? is it still needed? is there any Jira?
> 
> // this is a PR queue cleanup question

TL;DR: No Jira AFAIK, just proactive PR. Adds consistency in the CIV overall behavior. Not needed as you don't work with GCP (so far). Code works, but still issues with GCP rate limiting in case multiple instances are deployed in parallel.

Long story:
@alexxa , as far as you don't work on GCP then this can be closed. This PR attempted to fix parallel deployments (same as we do for AWS), however as GCP images don't behave the same as in AWS and it is more similar to Azure, it may not represent an issue. 

Anyways, this was also to have consistency across all the CIV supported cloud providers. Otherwise we should make a note in the readme.md explaining that parallel deployments in GCP is not working, only one instance at a time. 

The current code of this PR works as far as I remember, the only issue I had was about rate-limit from GCP so I was not able to spwan more than 5 instances simultaneously, if my memory does not fail.

I suggest you create the JIRA and link to this closed PR to have the reference for the future.

PR #112, opened 2022-08-24
Migrated kdump-related test cases, from os-tests.

narmaku:
Also changed the default instance type for AWS and GCP.
If the instance has 1GB or less, kdump serive fails to start
due to memory allocation issues.
Test names from os-tests:
- test_check_kdump_configuration
- test_check_kdump_status


narmaku:
The kdump tests are passing in the CI.

jabia99 (approved):
LGTM!

PR #113, opened 2022-08-24
CI: add aws.sh image builder test in CIV ci pipeline

jabia99:
CIV tool is being executed in aws.sh (IB CI), but it's running a specific release,
so there's a lot of manual work to keep this release updated. If we run
this test in out own CI and publish the container image, IB can always
get the last version of CIV without worrying if it's going to cause
errors.


.gitlab-ci.yml
```diff
@@ -10,6 +10,7 @@ stages:
     - sudo dnf update -y && sudo dnf install -y podman
   variables:
     RUNNER: aws/fedora-35-x86_64
+    INTERNAL_NETWORK: ["true"]
```
narmaku, line 13: Q: What is this env var used for?

jabia99: It allows the runners to be in the VPN

.gitlab-ci.yml
```diff
@@ -62,14 +64,47 @@ azure:
...
 
+.rhel_runners: &rhel_runners
+    RUNNER:
+      - aws/centos-stream-8-x86_64
+      - aws/rhel-8.4-ga-x86_64
+      - aws/rhel-8.6-ga-x86_64
+      - aws/rhel-9.0-ga-x86_64
+      - aws/rhel-8.7-nightly-x86_64
+      - aws/rhel-9.1-nightly-x86_64
+      - aws/centos-stream-9-x86_64
```
narmaku, line 83: I thick centos should be in .centos_runners

jabia99: centos is included in this category as it is considered a rhel like system, unlike fedora. I think the only reason this rhel-fedora separation is made in the yaml is to give acces to rhel like systems to the internal network. But we can separate centos

PR #116, opened 2022-08-30
use :latest tag instead of :main

jabia99:
To follow container registries conventions, we'll use :latest tag for
the last build instead of main.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -31,15 +31,12 @@ init:
   script:
     - schutzbot/update_github_status.sh start
 
-branch_container:
+container:
   extends: .deps
   stage: build
   script:
-    - sudo podman build -t civ:"${CI_COMMIT_REF_SLUG}" .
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -105,6 +105,16 @@ ib-aws:
     - schutzbot/deploy.sh
     - ARTIFACTS=/tmp/ bash /usr/libexec/tests/osbuild-composer/aws.sh
 
+container_main:
+  extends: .deps
+  rules:
```
narmaku, line 110: Maybe, instead of using rules, we can use the "only" section.

```
only:
    - main
```

narmaku: Another thing is, when the code is merged, the CI will look for `:main` container, and not `:latest`. So in that case we can add the "rule" section and declare a CI variable to use the correct tag.

jabia99: Only and except are not being actively developed, and might get deprecated: https://docs.gitlab.com/ee/ci/yaml/index.html#only--except

PR #147, opened 2022-10-04
test_suite: try to debug first_boot_time errors

jabia99:
this test is failing on aarch64 machines. print more information in test so we can debug what's happening


jabia99:
After some debugging. It seems there was a bug in the test code, where it interpreted `ms` as seconds

PR #151, opened 2022-10-11
Add Internal nightly runners

jabia99:


jabia99:
I checked that they are indeed the same internal nightlies as in osbuild-composer by temporarily re enabling the tests that were failing there, and checking that they indeed failed.

jrusz:
I've rebased the PR as there were a lot of weird errors, I'm hoping that will help.

jrusz (approved):
As for the PR I think it looks good, seems to be doing what it should, so LGTM. Let's just see if it actually passes.

jabia99:
> I've rebased the PR as there were a lot of weird errors, I'm hoping that will help.

I think most of the errors were infrastructure issues. (error against SP images are known and we are merging even if those jobs fail. Those jobs are called `azure` and `aws`). Let's merge when IB tests pass

PR #156, opened 2022-10-26
lib: exit on instance timeout

jabia99:
If the terraform controller timed out when trying to connect to the instances, it whould exit.


narmaku (approved):
LGTM!

PR #159, opened 2022-11-17
Added cloudx AWS account to the gitlab-ci file.

narmaku:
Testing with a marketplace AWS image, recently published.


jabia99:
LGTM! Cool that we finally have access to CloudX account

PR #163, opened 2022-11-25
ci: update terraform definitions to the latest ones

ondrejbudai:
Changes:
- Fedora 37 no longer requires any hacks
- All the images deleted by cloud-cleaner are back:
  - Centos Stream 9
  - All RHEL nightlies
- Instances shouldn't be killed as much anymore


jabia99:
Just FYI: Please ignore "aws" and "azure" jobs in CI. The only relevant ones right now are "IB-aws" and "IB-azure"

ondrejbudai:
azure passed though, is that bad? :stuck_out_tongue_winking_eye:

jabia99:
We are waiting until we can find what images we should test against. (As Alex would say, the latest and greatest). So for now we just ignore those tests. :)

Once a PR is merged, only IB tests run. If these pass, `prod` container image is updated

jabia99:
Thanks Ondrej, highly appreciate this :heart:

PR #168, opened 2022-12-14
Added developers pub keys to civ instances

sshmulev:
https://issues.redhat.com/browse/CLOUDX-244


jabia99 (commented):
Amazing! just a small comment.

Have you tried using the already existing keys file? this way we wouldn't have to maintain two files with the same content. Maybe instead of iterating each key, just echo the whole file into authorized_keys

jabia99 (approved):

PR #169, opened 2022-12-19
Add developers public keys to CIV instances

sshmulev:
https://issues.redhat.com/browse/CLOUDX-244


lib/ssh_lib.py
```diff
@@ -66,3 +67,35 @@ def copy_file_to_host(host, local_file_path, destination_path):
...
+
+    Args:
+        instances(dict): Has all the instances needed information - 'public_dns', 'username' etc..
+    """
+    authorized_keys = "~/.ssh/authorized_keys"
+    ssh_identity_file = '/tmp/ssh_key'
+    team_ssh_keys_file = 'schutzbot/team_ssh_keys.txt'
+    ssh_client = get_ssh_client()
+
+    for instance in instances.values():
```
narmaku, line 84: Challenge: How about doing this in parallel?
Hint: check `wait_for_all_instances_ssh_up` here:
https://github.com/osbuild/cloud-image-val/blob/3bd776f5afe5b6bda0a79be12cdd1535994684b4/cloud/terraform/terraform_controller.py#L30

lib/ssh_lib.py
```diff
@@ -66,3 +67,35 @@ def copy_file_to_host(host, local_file_path, destination_path):
 
     sftp.put(local_file_path, destination_path)
     sftp.close()
+
+
+def add_ssh_keys_to_instances(instances):
+    """
```
narmaku, line 73: For docstrings, let's stick to the same convention as we have in other parts of the code.
Example here:
https://github.com/osbuild/cloud-image-val/blob/3bd776f5afe5b6bda0a79be12cdd1535994684b4/lib/test_lib.py#L27

narmaku (changes requested):
Good job on this one! Just a couple of comments.

lib/ssh_lib.py
sshmulev: Ack

lib/ssh_lib.py
sshmulev: Seems like there is some difficulty with making Paramiko work with multithreaded connections.
Will track this if it causes timing issues in the future - If so, we'll have to think of another solution for that.

PR #178, opened 2023-01-12
CIV: Move reboot and boot tests to a separate test suite file

sshmulev:
https://issues.redhat.com/browse/CLOUDX-241

### DIFF
--- a/test_suite/generic/test_reboot.py
+++ b/test_suite/generic/test_reboot.py
@@ -1,0 +1,75 @@
+import pytest
+from lib import test_lib
+
+
+class TestsReboot:
+    hostname_before_reboot_file = '/var/hostname_before_reboot'
+    kmemleak_arg = 'kmemleak=on'
+
+    def setup_before_reboot(self, host):
+        with host.sudo():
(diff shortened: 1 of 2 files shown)


test_suite/generic/test_reboot.py
```diff
@@ -0,0 +1,76 @@
+import pytest
+from lib import test_lib
+
+
+@pytest.mark.skip("To be analize by CLOUDX-230")
```
narmaku, line 5: I think we can remove this skip marker. This way we will be able to run these test on-demand if needed in the future. Right now it won't affect the logic of CIV since the paths to the test suites that need to be run, are harcoded here:
https://github.com/osbuild/cloud-image-val/blob/3bd776f5afe5b6bda0a79be12cdd1535994684b4/test_suite/suite_runner.py#L80
(we need to change this in the near future though, so this can be an extra parameter of the tool)

narmaku (changes requested):
Nice! Just one comment and we are ready to go.

test_suite/generic/test_reboot.py
sshmulev: Ack

PR #180, opened 2023-01-17
Add retry to test jobs in gitlab ci

jabia99:
Some tests are failling due to temporary infrastructure issues, add one retry to every test job to improve developer experience


narmaku (approved):
Awesome! Easy as pie!

PR #186, opened 2023-01-29
CIV: Extend skip marker

sshmulev:
Many tests needed to be skipped by bigger or smaller versions, this should simplify this operation.
Another approach to this is to get the version release in the init part of the code and use the 'skipif' builtin fixture - think this would be a better way to to this.

### DIFF
--- a/test_suite/conftest.py
+++ b/test_suite/conftest.py
@@ -1,6 +1,7 @@
 import json
 import time
 import pytest
+import re
 
 from py.xml import html
 from pytest_html import extras
@@ -10,6 +11,11 @@ from lib import test_lib
 
 @pytest.fixture(autouse=True, scope='function')
 def check_markers(host, request):
+    skip_message = "This test doesn't apply to {distro_version}"
+
+    def parse_distro_version(distro_version):
+        return float(re.findall(r"\d+(\.\d+)?", distro_version)[0])
+
     # Check if test needs to wait before being run
     wait_marker = request.node.get_closest_marker('wait')
     if wait_marker:
@@ -32,7 +38,7 @@ def check_markers(host, request):
     ]
 
     host_distro = host.system_info.distribution
-    host_version = host.system_info.release
+    host_version = float(host.system_info.release)
     distro_version = f'{host_distro}{host_version}'
 
     # Skip the test if the distro is explicitly excluded
@@ -40,7 +46,19 @@ def check_markers(host, request):
         exclude_on_marker_list = exclude_on_marker.args[0]
         if host_distro in exclude_on_marker_list \
                 or distro_version in exclude_on_marker_list:
-            pytest.skip(f"This test doesn't apply to {distro_version}")
+            pytest.skip(skip_message.format(distro_version=distro_version))
+        else:
+            for item in exclude_on_marker_list:
+                if host_distro in item and item[0] == '<':
+                    if host_version < parse_distro_version(item):
+                        pytest.skip(skip_message.format(distro_version=distro_version))
+                    else:
+                        return
+                if host_distro in item and item[0] == '>':
+                    if host_version > parse_distro_version(item):
+                        pytest.skip(skip_message.format(distro_version=distro_version))
+                    else:
+                        return
 
     # If there is no run_on_marker and distro is not excluded execute supported
     # tests
@@ -62,7 +80,7 @@ def check_markers(host, request):
             or distro_version in run_on_marker_list:
         return
     else:
-        pytest.skip(f"This test doesn't apply to {distro_version}")
+        pytest.skip(skip_message.format(distro_version=distro_version))
 
 
 @pytest.mark.run_on(['rhel'])


jabia99 (approved):
Looks great! TY

jabia99:
I'm just thinking how we are going to do when a change is introduced in for example 8.8 and 9.2. If you put `>8.7` it will include 9.0 and 9.1

I guess we'll manually exclude those versions. I can't come up with a better solution. Maybe this condition should only apply inside the same major version??

sshmulev:
> I'm just thinking how we are going to do when a change is introduced in for example 8.8 and 9.2. If you put `>8.7` it will include 9.0 and 9.1
> 
> I guess we'll manually exclude those versions. I can't come up with a better solution. Maybe this condition should only apply inside the same major version??

I guess you meant if we need to exclude a spesicif version in addition? -  This feature still exists in line 41. The signs < or > are an option not a mandatory.

jabia99:
Let's have a meeting to discuss how to do this. I think it might be a bit more complex that expected

jabia99:
So after the last conversation, we agreed that we have two main options. One where the `>` or `<` character apply to all versions (`option 1`), and a second option where they apply only inside the same major version (`option 2`).

Let's imagine a scenario where we have to run a test from 8.8 and from 9.2. This means  this test should be executed  on `8.8`, `8.9`… and `9.2`, `9.3`..

For `option 1` we could use this marker

- `run_on([“>rhel8.7”, “>rhel9.1”])`

for `option 2` this would one way to do it

- `run_on([“>rhel8.7”])`
- `exclude_on([“rhel9.0”, “rhel9.1”])`

We can also imagine a second scenario were a test applies from 8.8 but doesn't apply to 9.x

`Option 1`

- `run_on([“<rhel9.0”])`
- `exclude_on([“<rhel8.8"])`

`Option 2`

- `run_on([“>rhel8.7”])`

Just chose the option that you find the best, we can always change it later if we see it requires a lot of maintenance :)

sshmulev:
Added another marker just for major versions 'major_version' - @pytest.mark.major_version(['rhel8.5', 'rhel8.6', 'rhel9.0'])
The marker will skip the versions beneath the major versions.
In my opinion, this marker needed to be separated from the other existing ones because it has a specific aim of use, unlike the 'exclude_on' marker that is more general in its range

test_suite/conftest.py
```diff
@@ -32,15 +39,36 @@ def check_markers(host, request):
...
+    host_version = float(host.system_info.release)
     distro_version = f'{host_distro}{host_version}'
 
     # Skip the test if the distro is explicitly excluded
     if exclude_on_marker:
         exclude_on_marker_list = exclude_on_marker.args[0]
         if host_distro in exclude_on_marker_list \
                 or distro_version in exclude_on_marker_list:
-            pytest.skip(f"This test doesn't apply to {distro_version}")
+            pytest.skip(skip_massage.format(distro_version=distro_version))
```
narmaku, line 50: I think there's a typo here. I would never skip a massage ;)

test_suite/conftest.py
```diff
@@ -32,15 +39,36 @@ def check_markers(host, request):
...
+
+        elif len(exclude_on_marker_list) > 1 and host_distro in exclude_on_marker_list[0]:
```
narmaku, line 52: Why are we checking if this list contains more than one item? Does this mean that it won'r work if we pass only one item? E.g. '>rhel9.0'

test_suite/conftest.py
```diff
@@ -32,15 +39,36 @@ def check_markers(host, request):
...
+                    pytest.skip(f"Skip test with version smaller than {exclude_on_marker_list[0]}")
+            elif exclude_on_marker_list[1] == '>':
+                if host_version > exclude_on_marker_list_version:
+                    pytest.skip(f"Skip test with version bigger than {exclude_on_marker_list[0]}")
+
+    if major_version_marker:
+        major_version_marker_list = major_version_marker.args[0]
+        for major_version in major_version_marker_list:
+            major_version_marker_list_version = parse_distro_version(major_version)
+            if host_version >= major_version_marker_list_version:
```
narmaku, line 65: I think this first "if" is redundant since the second one will do the check we need.

test_suite/conftest.py
```diff
@@ -32,15 +39,36 @@ def check_markers(host, request):
...
+                if int(host_version) == int(major_version_marker_list_version):
```
narmaku, line 66: This if can be inverted so that you don't need to use "else".
```suggestion
                if int(host_version) != int(major_version_marker_list_version):
                    pytest.skip(skip_message.format(distro_version=distro_version))
```

test_suite/conftest.py
```diff
@@ -10,6 +11,11 @@
 
 @pytest.fixture(autouse=True, scope='function')
 def check_markers(host, request):
+    skip_massage = "This test doesn't apply to {distro_version}"
+
+    def parse_distro_version(distro_version_list):
+        return float(re.findall(r"\d+\.\d+", distro_version_list)[0])
```
narmaku, line 17: This regex won't work in the case we want to filter only by distro major version
image
I think the purpose of the new marker "major_version" is to only add items such as "rhel8" or "rhel9".

test_suite/conftest.py
```diff
@@ -21,9 +27,10 @@ def check_markers(host, request):
     # If the host doesn't meet any of the combinations, the test will be skipped.
     run_on_marker = request.node.get_closest_marker('run_on')
     exclude_on_marker = request.node.get_closest_marker('exclude_on')
-    if not run_on_marker and not exclude_on_marker:
+    major_version_marker = request.node.get_closest_marker('major_version')
```
narmaku, line 30: I personally don't mind having this new marker for this kind of filtering. However we can only use it for inclusion and not for exclusion. The original idea was to have the capability to use major version filtering in the already existing markers (run_on and exclude_on). So, this would add the freedom of either add a specific minor version of a release, or just filter by major version, or a combination of both for the same marker.

(please note that it's not the same as filtering by "greater than" or "lower than", which should theoretically affect either major versions only or major and minor ones).

Since this is a temp solution until we find a better one e.g. not having to use markers as we discussed last week, I think we are OK for now.

test_suite/conftest.py
```diff
@@ -10,6 +11,11 @@
 
 @pytest.fixture(autouse=True, scope='function')
 def check_markers(host, request):
+    skip_massage = "This test doesn't apply to {distro_version}"
+
+    def parse_distro_version(distro_version_list):
```
narmaku, line 16: I think the name of this arg is misleading. We are not passing a list, but a single item which is a string containing distro, major version and minor version. E.g. "rhel9.1".

test_suite/conftest.py
narmaku: you can enhance the regex by making the .x minor version part optional: `\d+(.\d+)?`

test_suite/conftest.py
```diff
@@ -32,15 +39,36 @@ def check_markers(host, request):
...
+                if host_version < exclude_on_marker_list_version:
+                    pytest.skip(f"Skip test with version smaller than {exclude_on_marker_list[0]}")
+            elif exclude_on_marker_list[1] == '>':
+                if host_version > exclude_on_marker_list_version:
+                    pytest.skip(f"Skip test with version bigger than {exclude_on_marker_list[0]}")
+
+    if major_version_marker:
+        major_version_marker_list = major_version_marker.args[0]
+        for major_version in major_version_marker_list:
+            major_version_marker_list_version = parse_distro_version(major_version)
```
narmaku, line 64: I find these names a bit confusing, the combination of marker name + list + version makes a bit difficult to follow the code while reading it. It is also because there are other variables named similarly, such as the one above this line: `major_version_marker_list`.

I would suggest to use simpler names and try not to use "version" twice.

narmaku (changes requested):
Some comments added.

test_suite/conftest.py
sshmulev: The 'skip_message' is just a string, I noticed that it is repeated many times in the code so I think it is better to set one parameter for that.
Since this is the comment of the"pytest.skip", I thought that parameter named "skip_message" will define the operation - may another name be more appropriate?
skip_massage = "This test doesn't apply to {distro_version}"

test_suite/conftest.py
sshmulev: I forgot to mention how to use this, so here how it supposed to be used:
Until now we used it like this:
@pytest.mark.exclude_on(['rhel7.9'])
@pytest.mark.exclude_on(['rhel9'])
and so on...

With the fix you can still use the previous way to exclude a specific version or to set a sign '<' or '>' to refer the version you put in the fixrt index of the string.
@pytest.mark.exclude_on(['rhel7.9', '<'])
@pytest.mark.exclude_on(['rhel8.5', '>'])
I f that doesn't make sense to you we can think of something else together

test_suite/conftest.py
sshmulev: The first if checks if the version is bigger the the "major version" we compare, since the versions are float we need to make so that this scenario for example will succeed as well:
major version = 8.5
host version = 9.0
In this case we want to skip this test for host version with 9.0 because this is not as part of the major release, we should run only tests 8.5,8.6,8.7,8.8,8.9
The second skip is necessary for this scenario

In addition, if we leave only the second if we will fail also in this scenario:
major version = 8.5
host version = 8.3
It will not skip the version 8.3 even though it should because both int are 8

test_suite/conftest.py
sshmulev: This is a possible way, although I thought about something else as well:
            if host_version >= major_version_marker_list_version \
                    and int(host_version) == int(major_version_marker_list_version):
                return
            else:
                pytest.skip(skip_massage.format(distro_version=distro_version))

what do you think?

test_suite/conftest.py
sshmulev: ACK

sshmulev: regarding "I think the purpose of the new marker "major_version" is to only add items such as "rhel8" or "rhel9". Or even fedora3x as we have to test fedora sometimes too."
>>  Juan said that major releases could be 8.6 ,9.6 for example,in order to run tests bigger than them
^^@jabia99

test_suite/conftest.py
sshmulev: Yes you are right, this new marker is useful for inclusion, I thought it would be better to separate it since @jabia99 raised this issue with the major realses that we also might want to have by control when we run tests.
I thought it was better to separate this case for exclution because the regular exclusion is more generic, sometime you'd like to exclude all version beneath a specific one. but with this Major release is more limited

test_suite/conftest.py
narmaku: Sorry if I did not explain myself clear enough. My message was only about the fact that the variable has a typo and it is called "skip_massage" instead of "skip_message".

test_suite/conftest.py
narmaku: Then the name of the marker is misleading or it has more than one purpose. If it's called "major_version" I would think or assume that it is only for filtering by major versions here (I bet that other contributors may reach to the same conclusion at first sight). Why would this marker need to check for minor versions as well? I think having a separate marker for this serves also for the purpose of separating the logic that is only used for major version checks.

test_suite/conftest.py
narmaku: Oh, thanks. Now I understand by you check if there is more than 1 item in the list! I would invert the order anyways, so it is more readable and it looks more natural.
E.g. `@pytest.mark.exclude_on(['<', 'rhel8.8'])` (so, all versions lower than RHEL 8.8 will be skipped).

test_suite/conftest.py
sshmulev: You're right:)
will fix

test_suite/conftest.py
narmaku: Yeah I was thinking about the same too! However let's see what's the outcome of this comment:
https://github.com/osbuild/cloud-image-val/pull/186#discussion_r1097132205

test_suite/conftest.py
sshmulev: ACK

test_suite/conftest.py
sshmulev: got you:)
fixing it

narmaku: Thanks I just found the typo funny in this case because it's like skipping a massage session (which I would not skip because I love massages) LOL.

narmaku (approved):
Looking good!

jabia99 (approved):
Cool!!

PR #187, opened 2023-02-01
Add image gallery support to Azure VHD test

FKolwa:
This PR adds v2 of the Azure config builder.
- Added cloudx_image_gallery on our Azure account.
- Added architecture property to instance config (important for non vhd image tests)
- Added default VM size for arm64 tests.
- Added logic to parse necessary config information from VHD blob name.
- Added logic to create new shared image versions on demand.

Uncertainties:
It seem that Terraform does not require us to manually create shared image definitions. The logic for creating new image definitions is included, but the execution has been commented out.


cloud/terraform/azure_config_builder_v2.py
```diff
@@ -0,0 +1,332 @@
...
+
+        target_region = {
+            'name': 'eastus',
+            'regional_replica_count': 1
+        }
+
+        new_image = {
+            'name': '0.0.1',
+            'blob_uri': instance['vhd_uri'],
+            'storage_account_id': f"/subscriptions/{self.subscription_id}/resourceGroups/rh-resource/providers/Microsoft.Storage/storageAccounts/rhimages",
```
narmaku, line 153: Is there a way that we can compose this URI without having to hardcode the storage account and other things?

cloud/terraform/azure_config_builder_v2.py
```diff
@@ -0,0 +1,332 @@
...
+                new_instance['plan'] = instance['plan']
+        elif 'vhd_uri' in instance:
+            new_instance['depends_on'].append('azurerm_shared_image_version.{}'.format(instance['azurerm_shared_image_version']))
+            new_instance['source_image_id'] = self.__get_azure_image_uri(instance['azurerm_shared_image'])
+
+        self.resources_tf['resource']['azurerm_linux_virtual_machine'][instance_hostname] = new_instance
+
+    def __get_azure_image_uri(self, azure_image_name):
+        """
+        Returns a composed string URI that belongs to a specific Azure image, from its name
```
narmaku, line 300: I would specify that this related to a shared image in an Image Gallery.

cloud/terraform/terraform_configurator.py
```diff
@@ -53,7 +53,7 @@ def get_config_builder(self):
         if cloud_name == 'aws':
             return AWSConfigBuilder(self.resources_dict, self.ssh_key_path)
         elif cloud_name == 'azure':
-            return AzureConfigBuilder(self.resources_dict, self.ssh_key_path)
+            return AzureConfigBuilderV2(self.resources_dict, self.ssh_key_path)
```
narmaku, line 56: By replacing this V1 version by the V2, we are braking current functionality which is e.g. running tests on images that come from the marketplace (see `cloud/sample/resources_azure_marketplace.json` and `cloud/sample/resources_azure.json`). so, we either implement a factory for this Azure config builders, or directly check if the `vhd_uri` element is present in the json, to chose to use the V2 one. Any other options/ideas?

cloud/terraform/azure_config_builder_v2.py
```diff
@@ -0,0 +1,332 @@
...
+            ]
+        }
+
+        self.resources_tf['resource']['azurerm_network_interface'][name] = new_nic
+
+    def __new_azure_vm(self, instance):
+        if 'instance_type' not in instance or not instance['instance_type']:
+            if instance['arch'] == 'x86_64':
+                instance['instance_type'] = self.default_x86_vm_size
+            elif instance['arch'] == 'arm64':
```
narmaku, line 247: I would add "else" here. Sometimes we don't specify the VM size and the default one should be used (assuming that it's x64 arch - we do the same for AWS).

cloud/terraform/azure_config_builder_v2.py
FKolwa: Hm. Technically we could get all the information we need to build any image from the VHD uri, so yes we could extract it as part of the image name parsing as well. Otherwise we would have to set that as part of the parameters for the pipeline request. What do you think?

cloud/terraform/azure_config_builder_v2.py
FKolwa: :+1:

jabia99:
Before merging this, we have to update cloud cleaner to remove the image galleries, image definitions and image versions.

FKolwa:
> Before merging this, we have to update cloud cleaner to remove the image galleries, image definitions and image versions.

Good call @jabia99 ! I'd say we should skip the image gallery though. It's a resource group and does not create any costs. We reuse the same image gallery for all our tests / images. Image definitions and versions however should be cleaned, especially because those resources have to be unique and will cause tests to fail if you try to overwrite them.

narmaku (approved):

PR #189, opened 2023-02-03
Fix civ network manager cloud setup

narmaku:


jabia99 (commented):
Shouldn't we wait for https://github.com/osbuild/cloud-image-val/pull/186 and just do 
`@pytest.mark.exclude_on(['<rhel8.5')`

jabia99:
I'm guessing the only relevant commit is the last one, right? I see 28 commits, guessing it's a branch issue

narmaku:
> Shouldn't we wait for #186 and just do `@pytest.mark.exclude_on(['<rhel8.5')`

Correct. Once that PR is merged, we will be able to remove this check from the source code and update the marker.
https://github.com/osbuild/cloud-image-val/pull/189/commits/cd447b23d23f0641e42cc424122a4d2e8291c25e#diff-6e493f324d72e518b7e1570ff75ebdad6f8125756bd4e28e6bbc9cb7f413cbabR608

narmaku:
> I'm guessing the only relevant commit is the last one, right? I see 28 commits, guessing it's a branch issue

Yes, sorry about that. I have no idea why this happens sometimes. I guess it must have something to do with the rebase that is not done in the correct order with the commit.

narmaku:
Ok, so this PR looks ugly with that many non-sense commits that are already in main. I will close this PR and create a new one.

PR #190, opened 2023-02-06
Fix test_redhat_cds_hostnames test case.

narmaku:
In some AWS regions, some CDS hosts are not available. Also, there are others that have connectivity issues or simply are being deprecated (e.g. rhui2). So, the logic was enhanced to check first the newest ones and print the ones that are failing to communicate with the instance. Lastly, the logic checks all of the CDS and the test now fails only if ALL of them are unavailable.

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -749,11 +749,14 @@ class TestsAWSNetworking:
         region = instance_data_aws_web['region']
 
         rhui_cds_hostnames = [
-            f'rhui2-cds01.{region}.aws.ce.redhat.com',
-            f'rhui2-cds02.{region}.aws.ce.redhat.com',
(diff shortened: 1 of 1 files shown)


jabia99 (approved):
LGTM

jabia99:
Just a question, how often do these CDS hosts get updated? Just to know how often we should maintain this test

sshmulev (approved):
+1

narmaku:
> Just a question, how often do these CDS hosts get updated? Just to know how often we should maintain this test

I just got a reply internally and this should not change to often. Maybe every a couple of years or so. It will depend on the life-cycle of the rhui versions that can be found in the produce page here: https://access.redhat.com/support/policy/updates/rhui

Also, the newly introduced one, which is rhui.{region} (it is un-versioned) should always work form now on, as it automatically redirects to the proper one.

PR #191, opened 2023-02-07
Fix test_kdump_status test case.

narmaku:
This test case is sometimes failing due to lack of RAM in the running instance. By setting an instance type with more resources, this should be fixed. This test also indirectly fixed test_no_fail_service test case, since this latter one is also failing when kdump did not work as expected.

### DIFF
--- a/cloud/terraform/aws_config_builder.py
+++ b/cloud/terraform/aws_config_builder.py
@@ -55,7 +55,9 @@ class AWSConfigBuilder(BaseConfigBuilder):
 
     def __new_aws_instance(self, instance):
         if not instance['instance_type']:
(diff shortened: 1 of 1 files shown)


jabia99 (approved):
LGTM! Let's hope this was the root cause, it is strange that it was failing only on some instances and not all.

narmaku:
> LGTM! Let's hope this was the root cause, it is strange that it was failing only on some instances and not all.

Thanks! Yes, this should solve the issues. If the issue re-appears, we can be sure it's a service issue and not because of lack of RAM. This was confirmed by Frank.

sshmulev (approved):
+1

PR #192, opened 2023-02-07
Fix test_network_manager_cloud_setup

narmaku:
This test case was being run on RHEL lower than 8.4 since the exclusion logic was only applying to RHEL 8.4 This needs to be excluded in ANY RHEL version lover than 8.5.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -586,7 +586,7 @@ class TestsNetworking:
                 f'Unexpected device name. Expected: "{device_name}"'
 
     @pytest.mark.run_on(['rhel'])
-    @pytest.mark.exclude_on(['rhel7.9', 'rhel8.4'])
+    @pytest.mark.exclude_on(['<rhel8.5'])
     def test_network_manager_cloud_setup(self, host, instance_data):
         """
         BugZilla 1822853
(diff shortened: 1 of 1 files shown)


jabia99 (approved):
LGTM!

test_suite/generic/test_generic.py
```diff
@@ -586,7 +586,7 @@ def test_eth0_network_adapter_setup(self, host):
                 f'Unexpected device name. Expected: "{device_name}"'
 
     @pytest.mark.run_on(['rhel'])
-    @pytest.mark.exclude_on(['rhel7.9', 'rhel8.4'])
+    @pytest.mark.exclude_on(['<rhel8.5'])
```
sshmulev, line 589: seems like the test "test_network_manager_cloud_setup" passed on rhel-8.4 -  but it should be skipped no?

sshmulev: I noticed to an issue with the regex of the  exclude_on marker will create a PR to fix it - this is why it didn't skip

sshmulev: https://github.com/osbuild/cloud-image-val/pull/195

sshmulev (approved):

PR #194, opened 2023-02-07
CIV: Update readme file

sshmulev:
A new feature for 'exclude_on' marker has been added which helps to skip versions in a range under or bigger then a specified version, with a help of the opeartors ('<' or '>').


narmaku (approved):
Approved!

PR #195, opened 2023-02-08
CIV: fix exclude_on marker

sshmulev:
regex was incorrect

### DIFF
--- a/test_suite/conftest.py
+++ b/test_suite/conftest.py
@@ -14,7 +14,7 @@ def check_markers(host, request):
     skip_message = "This test doesn't apply to {distro_version}"
 
     def parse_distro_version(distro_version):
-        return float(re.findall(r"\d+(\.\d+)?", distro_version)[0])
+        return float((re.search(r"\d+(\.\d+)?", distro_version)).group(0))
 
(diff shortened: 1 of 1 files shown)


test_suite/conftest.py
```diff
@@ -14,7 +14,7 @@ def check_markers(host, request):
     skip_message = "This test doesn't apply to {distro_version}"
 
     def parse_distro_version(distro_version):
-        return float(re.findall(r"\d+(\.\d+)?", distro_version)[0])
+        return float((re.search(r"\d+(\.\d+)?", distro_version)).group(0))
```
narmaku, line 17: Could this bring issues when there's actually only major version provided?
image
E.g. for fedora, which does not include minor version.

sshmulev: This won't raise issues:
If we want to exclude rhel or fedora35 distro for example you don't need to use the operators < >, so this function won't be call.

narmaku: You're right! Thanks for the great explanation. In that case all good from my side about this PR :+1:

narmaku (approved):

PR #196, opened 2023-02-09
Add a JSON report as part of the artifacts generated by the tool.

narmaku:
This is a previous step to unblock other use cases to process test report data such as the CIV Report Analyzer tool we are developing right now.


jabia99 (approved):
LGTM!

PR #209, opened 2023-02-27
Added CIV Report Analyer tool.

narmaku:
This is to give a brief summary and insights of the report generated after running tests. This analyzer tool digests any report.json generated by CIV after running tests on cloud instances. It is an independent tool and can be used after running cloud-image-val main script.


jabia99 (approved):
LGTM!!

PR #210, opened 2023-03-01
remove test_kdump_conf

jabia99:
this test was not adding a lot of value and it needed a maintenance all the time. So it gets removed so it doesn't interfere with validations

Context: CLOUDX-202


narmaku (approved):
Approved!

PR #228, opened 2023-04-26
Modify timeout for ssh

jabia99:
Since adding ap-southeast-4 regions, those instances are timing out. Trying to see if this is the reason


jabia99:
Canceling CI in order to save resources.

PR #234, opened 2023-05-03
Reenable test in rhel 8.4

ghost:
Add context to test and reenable it


narmaku (approved):
Looks great!

narmaku (approved):
Looks great!

jabia99:
This is failing still failing in 8.4. But my understanding from the bugzilla is that is shouldn't fail...

PR #239, opened 2023-05-18
CIV Report Analyzer improvements. 

narmaku:
Enhanced the analyzer "table" output option to print more fields that fit into the "Execution Analysis" spreadsheet.

### DIFF
--- a/main/civ_report_analyzer.py
+++ b/main/civ_report_analyzer.py
@@ -43,39 +43,49 @@ def get_failed_tests_analysis(data):
     return analysis
 
 
-def get_formatted_analysis(analysis, format):
+def get_formatted_analysis(report_data, format):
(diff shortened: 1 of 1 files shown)


main/civ_report_analyzer.py
```diff
@@ -23,6 +23,26 @@
 spaced_indentation = ' ' * 8
 
 
+def get_failed_tests_analysis_2(data):
```
jabia99, line 26: Is this function left from debugging? It seems to be a duplicate

jabia99 (commented):
Looks good! just a comment

main/civ_report_analyzer.py
narmaku: Good catch! My mistake. I'll remove it!

PR #249, opened 2023-06-08
Fixes for Fedora 38

ondrejbudai:
Two adjustments for the tests to pass on Fedora 38 images. See individual commits and comments inline for more info.


test_suite/cloud/test_aws.py
```diff
@@ -355,12 +355,20 @@ def test_hostkey_permissions(self, host):
         """
         Check that ssh files permission set are correct.
         BugZilla 2013644
+
+        Excluded on Fedora >= 38 because the
```
jabia99, line 359: Unfinished sentence?

test_suite/cloud/test_aws.py
```diff
@@ -355,12 +355,20 @@ def test_hostkey_permissions(self, host):
...
         """
         files_to_check = ['ssh_host_ecdsa_key',
                           'ssh_host_ed25519_key', 'ssh_host_rsa_key']
+        expected_mode = 0o640
+        if host.system_info.distribution == 'fedora' and int(host.system_info.release) >= 38:
```
jabia99, line 364: Will this change get into RHEL? If so, can we already know which version will include this change?

test_suite/generic/test_generic.py
```diff
@@ -344,8 +344,14 @@ def test_sshd(self, host):
                 pytest.fail('ssh.service is not active')
 
             pass_auth_config_name = 'PasswordAuthentication'
+            pass_auth_config_file = '/etc/ssh/sshd_config'
 
-            if not host.file('/etc/ssh/sshd_config').contains(f'^{pass_auth_config_name} no'):
+            if host.system_info.distribution == 'fedora' and int(host.system_info.release) >= 38:
```
jabia99, line 349: Same as above, will this changes get into RHEL?

jabia99 (commented):
Thanks a lot! Just a few questions

test_suite/cloud/test_aws.py
ondrejbudai: My wild guess: Red Hat didn't rebase openssh during the whole lifecycle of RHEL 8, thus I don't expect this to happen neither in RHEL 9 unless someone explicitly requests it. I don't think that's quite unlikely, though. So my guess is it will appear in RHEL 10.

test_suite/generic/test_generic.py
ondrejbudai: Red Hat rebases cloud-init in RHEL, thus it might appear at some point in 8.x and 9.x. When exactly is a question for the maintainers of cloud-init, though.

jabia99 (approved):

PR #238, opened 2023-05-17
Add 8.9 and 9.3 runners

ghost:

PR #250, opened 2023-06-12
WIP - sshd_config and smart_auth test cases

narmaku:
- Fixed test_sshd_config test case.


atodorov:
FTR can be rebased on top of #238

test_suite/generic/test_generic.py
```diff
@@ -345,7 +363,7 @@ def test_sshd(self, host):
 
             pass_auth_config_name = 'PasswordAuthentication'
 
-            if not host.file('/etc/ssh/sshd_config').contains(f'^{pass_auth_config_name} no'):
+            if not host.file(sshd_config_file).contains(f'^{pass_auth_config_name} no'):
                 pytest.fail(f'{pass_auth_config_name} should be disabled')
```
atodorov, line 367: I don't think this will work. 

There's also the file `/etc/ssh/sshd_config.d/50-redhat.conf` which contains `ChallengeResponseAuthentication no`. That is also another way to control password authentication with sshd. Here's how the 2 files look in combination for me:

```
# cat sshd_config | grep Password
#PasswordAuthentication yes
#PermitEmptyPasswords no
# PasswordAuthentication.  Depending on your PAM configuration,
# PAM authentication, then enable this but set PasswordAuthentication


# cat sshd_config.d/50-redhat.conf 
# This system is following system-wide crypto policy. The changes to
# crypto properties (Ciphers, MACs, ...) will not have any effect in
# this or following included files. To override some configuration option,
# write it before this block or include it before this file.
# Please, see manual pages for update-crypto-policies(8) and sshd_config(5).
Include /etc/crypto-policies/back-ends/opensshserver.config

SyslogFacility AUTHPRIV

ChallengeResponseAuthentication no

GSSAPIAuthentication yes
GSSAPICleanupCredentials no

UsePAM yes

X11Forwarding yes

# It is recommended to use pam_motd in /etc/pam.d/sshd instead of PrintMotd,
# as it is more configurable and versatile than the built-in version.
PrintMotd no

# Accept locale-related environment variables
AcceptEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
AcceptEnv LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT
AcceptEnv LC_IDENTIFICATION LC_ALL LANGUAGE
AcceptEnv XMODIFIERS
```

PR #259, opened 2023-07-11
Fixed test_sshd test case which was failing in RHEL 9.3 and 8.9.

narmaku:
Also added more context to the test_sshd test case stdout.
About test_smartcard_auth test case, it will now be skipped until a Bugzilla is fixed and a release is done for that, affecting RHEL 8.9.


atodorov (approved):
LGTM

PR #266, opened 2023-08-09
Add a test to assert on /etc/machine-id permissions for AWS

atodorov:
Related: RHBZ #2221269

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -29,6 +29,14 @@ def instance_data_aws_cli(host, instance_data_aws_web):
 
 @pytest.mark.order(2)
 class TestsAWS:
+    @pytest.mark.run_on(['rhel', 'fedora'])
+    def test_etc_machine_id_permissions(self, host, instance_data):
+        """
+        Check that /etc/machine-id permissions are 444.
+        Bugzilla: 2221269
+        """
+        assert host.file("/etc/machine-id").mode == 0o444, "Expected 444 permissions for /etc/machine-id"
+
     # TODO: Divide test. Analyze centos
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel', 'fedora'])


atodorov:
@narmaku can you tag a new container version once this is approved and merged ? See https://bugzilla.redhat.com/show_bug.cgi?id=2221269#c6

test_suite/cloud/test_aws.py
```diff
@@ -29,6 +29,22 @@ def instance_data_aws_cli(host, instance_data_aws_web):
 
 @pytest.mark.order(2)
 class TestsAWS:
+    @pytest.mark.pub
+    @pytest.mark.run_on(['rhel', 'fedora'])
+    def test_etc_machine_id_permissions(self, host, instance_data):
+        """
+        Check that /etc/machine-id permissions are 444. RHBZ #2221269
```
narmaku, line 36:
```suggestion
        Check that /etc/machine-id permissions are 444.
        Bugzilla: 2221269
```

narmaku, line 36:
```suggestion
        Check that /etc/machine-id permissions are 444.
        Bugzilla: 2221269
```

test_suite/cloud/test_aws.py
```diff
@@ -29,6 +29,22 @@ def instance_data_aws_cli(host, instance_data_aws_web):
...
+        """
+        machine_id = "/etc/machine-id"
+
+        assert os.access(machine_id, os.R_OK), f"Expected read access for {machine_id}"
```
narmaku, line 40: Maybe this can be simplified to just check that the octal permissions are `444`, using testinfra's host module.
I think that os.* will not be run in the remote host but on the runner itself (e.g. locally in Jenkins agent or GitLab runner). For that we need to use `host.run('whatever shell cmd')` or similar.

See this test case as an example:
* https://github.com/osbuild/cloud-image-val/blob/fcc748864776b1363cc165bdfc7b520c360cb35e/test_suite/cloud/test_aws.py#L378C75-L378C75

You can find more information and usages of testinfra modules here:
https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.file.File.mode

narmaku: About execution access and similar, we may have to run shell commands with `host.run()` and capture `stdout` from it.

narmaku (commented):
Let's use testinfra modules for this. It's the pytest library we use to run tests in remote hosts. Let me know if you need help with that.

narmaku (approved):
LGTM!

PR #278, opened 2023-09-21
CIV: fix "test_network_ipv6_setup"

sshmulev:


narmaku:
@sshmulev did you re-test with AMIs where the test ran and actually had the failure? IB tests from the CI are being skipped.
Please note that the logic of the test case skips it if ipv6 is not enabled in the subnet.
So, in this case, if the test is being skipped most of the times, we cannot verify that this fix is working or not.
See for example: https://redhat.gitlab.io/-/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/5135428442/artifacts/report.html

narmaku (commented):
Let's run this test in real-life AMIs where we encountered failures.

sshmulev:
> @sshmulev did you re-test with AMIs where the test ran and actually had the failure? IB tests from the CI are being skipped. Please note that the logic of the test case skips it if ipv6 is not enabled in the subnet. So, in this case, if the test is being skipped most of the times, we cannot verify that this fix is working or not. See for example: https://redhat.gitlab.io/-/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/5135428442/artifacts/report.html

Sure I tested it on AMIs that failed, and after this change they passed. I verified it with several AMIS.

narmaku:
> > @sshmulev did you re-test with AMIs where the test ran and actually had the failure? IB tests from the CI are being skipped. Please note that the logic of the test case skips it if ipv6 is not enabled in the subnet. So, in this case, if the test is being skipped most of the times, we cannot verify that this fix is working or not. See for example: https://redhat.gitlab.io/-/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/5135428442/artifacts/report.html
> 
> Sure I tested it on AMIs that failed, and after this change they passed. I verified it with several AMIS.

Thanks for the confirmation! Let's merge this!

PR #279, opened 2023-09-27
CIV: debug "test_yum_group_install"

sshmulev:


test_suite/cloud/test_aws.py
```diff
@@ -552,13 +552,27 @@ def test_yum_group_install(self, host):
...
-        with host.sudo():
-            assert host.run_test('yum -y groupinstall "Development tools"'), \
-                'Error while installing Development tools group'
-
-            package_to_check = 'glibc-devel'
-            assert host.package(package_to_check).is_installed, \
-                f'{package_to_check} is not installed'
+        max_attempts = 2
+        package_to_check = 'glibc-devel'
+        for attempt in range(1, max_attempts + 1):
```
narmaku, line 557: There is a retry mechanism already included in CIV thanks to the pytest-retry.
We have defined a list of error messages to "capture" by the retry mechanism so that it can re-ruin the test when encountering any of these failures:
https://github.com/osbuild/cloud-image-val/blob/a93d3c4d403707f664f3cdc375be5b303f031546/test_suite/suite_runner.py#L8
I think it will be be much easier to leave the current test case implementation as it is and just add the following error message to `rerun_failing_tests_regex`: "Error while installing Development tools group"
I think it's a great idea though to introduce some safety measurements as you commented in the associated JIRA, especially about cache metadata.

narmaku (changes requested):

test_suite/cloud/test_aws.py
sshmulev: ACK

narmaku (approved):

PR #280, opened 2023-09-28
A workaround for "yum and rpm get a db5 error"

sshmulev:


test_suite/cloud/test_aws.py
```diff
@@ -301,6 +301,10 @@ def test_required_packages_are_installed(self, host):
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel'])
     def test_rhui_pkg_is_installed(self, host):
+        with host.sudo():
+            # Workaround for "yum and rpm get a db5 error"
```
narmaku, line 305:
```suggestion
            # CLOUDX-590
```

test_suite/cloud/test_aws.py
```diff
@@ -301,6 +301,10 @@ def test_required_packages_are_installed(self, host):
...
+            assert host.run('rm -f /var/lib/rpm/__db.*').succeeded, 'failed to remove the RPM db files'
+            assert host.run('rpm --rebuilddb').succeeded, 'failed to rebuild the RPM db files'
```
narmaku, line 307: You can do this instead, to directly check that command did not fail:
```suggestion
            assert host.run_test('rpm --rebuilddb'), 'Failed to rebuild the RPM db files'
```
(same for the line above)
More info here: https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.host.Host.run_test

narmaku, line 307: I'm, thinking if we really want to stop test case execution here and make it fail... maybe we can just print a message about what went wrong at this stage but continue trying to check for the RPMs?

narmaku (changes requested):

test_suite/cloud/test_aws.py
sshmulev: ACK

test_suite/cloud/test_aws.py
sshmulev: ACK

sshmulev: I would keep the assert here for some reasons:
1. If this is actually the case that the DB is broken and this is why this test fails - then it is not really necessary to continue cause it will fail anyway.
2. If we find out that the assert works fine but still have issues with retrieving data from the DB we'll probably need to change the solution for it

narmaku: Thanks @sshmulev , makes total sense!

narmaku (approved):
LGTM!

PR #282, opened 2023-10-30
Added support to specify HyperV generation in resources.json.

narmaku:
Default will be "V2" if no value is provided or if it is null/empty.
Also improved VM name based on image name when it is available.

### DIFF
--- a/cloud/terraform/azure_config_builder_v2.py
+++ b/cloud/terraform/azure_config_builder_v2.py
@@ -9,6 +9,7 @@ class AzureConfigBuilderV2(BaseConfigBuilder):
 
     default_x86_vm_size = 'Standard_DS1_v2'
     default_arm64_vm_size = 'Standard_D2pls_v5'
+    default_hyper_v_generation = 'V2'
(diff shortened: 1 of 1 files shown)


cloud/terraform/azure_config_builder_v2.py
```diff
@@ -150,19 +160,25 @@ def __new_azure_shared_image(self, instance):
             'sku': version
         }
 
+        if 'hyper_v_generation' in instance and str(instance['hyper_v_generation']) != '':
```
atodorov, line 163: FTR ` and str(instance['hyper_v_generation']) != '':` is redundant. In a boolean context the values `None` and `""` (empty string) are evaluated as `False`. The whole condition should actually be 

```
        if 'hyper_v_generation' in instance and instance['hyper_v_generation']:
```

To avoid the if-else you can also use `getattr(instance, 'hyper_v_generation', self.default_hyper_v_generation)`.

atodorov (approved):
Added a small comment, otherwise looks good.

PR #286, opened 2023-11-07
Fixed version comparisons in all the test suites and conftest.py

narmaku:
Before we were comparing floats, but it turns out that it is not accurate, since in the case of RHEL 8.10, it is semantically greater than e.g. 8.4, however if they are compared as float, 8.4 is greater than 8.10.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 PyYAML==5.3.1
-pytest==7.4.0
+pytest~=7.4.3
 pytest-mock==3.10.0
 pytest-cov==4.0.0
 pytest-testinfra==8.1.0
@@ -15,3 +15,4 @@ py==1.11.0
 paramiko==3.1.0
 awscli==1.27.119
 requests==2.28.1
+packaging==23.2
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -1,6 +1,8 @@
 import json
 import re
+
 import pytest
+from packaging import version
 
 from lib import test_lib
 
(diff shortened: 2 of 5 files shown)


jrusz (approved):
LGTM! Let's see if CI agrees :) ...  Do you want me to test it in osbuild-composer test suite as well? It's just a matter of changing the CIV image tag, righ?

narmaku:
> LGTM! Let's see if CI agrees :) ... Do you want me to test it in osbuild-composer test suite as well? It's just a matter of changing the CIV image tag, righ?

Thanks @jrusz ! Yes, that would be great, so we can confirm that this also solves the issues with RHEL-8.10.
Exactly, you can use this PR as the tag (pr-286), instead of "prod" here:

- https://github.com/osbuild/osbuild-composer/blob/3fe36d00120d69907677b5a922650a14ff235ebd/test/cases/aws.sh#L223

jrusz:
It passed on both 9.4 and 8.10! https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1063680945

sshmulev (commented):
I would try to keep on a consistent argument  when comparing the versions:
right now I see 'rhel_version', 'product_version', 'version.parse(host.system_info.release)', 'product_release_version', maybe lets choose one argument for all these comparisons?

narmaku:
> I would try to keep on a consistent argument name when comparing the versions: right now I see 'rhel_version', 'product_version', 'version.parse(host.system_info.release)', 'product_release_version', maybe lets choose one argument for all these comparisons?

Good point. Maybe we can choose a generic one, like `system_release`. What do you think?

PR #288, opened 2023-11-20
CIV: fix for audit.conf test

sshmulev:
Need to add a new condition for both 9.4 and 8.10 distros with new hashes.

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -127,7 +127,7 @@ class TestsAWS:
         system_release = version.parse(host.system_info.release)
         if system_release >= version.parse('9.4'):
             checksums = checksums_by_version['9.4+']
-        elif system_release >= version.parse('8.10'):
+        elif version.parse('9.0') > system_release >= version.parse('8.10'):
             checksums = checksums_by_version['8.10+']
         elif system_release >= version.parse('8.6'):
             checksums = checksums_by_version['8.6+']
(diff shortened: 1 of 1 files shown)


jrusz:
Ok it passes in my test pipeline https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1078875121 you just need to update the snapshots for the image build to `20231115`

jrusz:
> Ok it passes in my test pipeline https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1078875121 you just need to update the snapshots for the image build to `20231115`

But it seems those are taken from osbuild-composer repository... If I update them it's going to be failing without this fix being merged but without updating them this fix is going to be failing, not sure how to proceed :sweat_smile:

test_suite/cloud/test_aws.py
```diff
@@ -117,7 +125,11 @@ def test_auditd(self, host):
             auditd_service).is_running, f'{auditd_service} expected to be running'
 
         system_release = version.parse(host.system_info.release)
-        if system_release >= version.parse('8.6'):
+        if system_release >= version.parse('9.4'):
+            checksums = checksums_by_version['9.4+']
+        elif system_release >= version.parse('8.10'):
```
jrusz, line 130: We'll need to add something like 
```suggestion
        elif system_release >= version.parse('8.10') and system_release < version.parse('9.0'):
``` 
(syntax is probably wrong) because otherwise it's matching the 9.1 GA which is not what we want, we want just rhel-8 matched.

jrusz:
The pipelines finally passed and so did the new nightlies, however there is one issue with rhel-9 ga.

jrusz:
I've added a commit to update the condition and I'm going to merge this, hope it's ok.

PR #289, opened 2024-01-08
CIV: Refactor test_rhui_pkg_is_installed

sshmulev:
https://issues.redhat.com/browse/CLOUDX-761


test_suite/cloud/test_aws.py
```diff
@@ -317,8 +317,17 @@ def test_required_packages_are_installed(self, host):
     def test_rhui_pkg_is_installed(self, host):
         with host.sudo():
             # CLOUDX-590
-            assert host.run_test('rm -f /var/lib/rpm/__db.*'), 'Failed to remove the RPM db files'
-            assert host.run_test('rpm --rebuilddb'), 'Failed to rebuild the RPM db files'
+            try:
+                assert host.run_test('rm -f /var/lib/rpm/__db.*'), 'Failed to remove the RPM db files'
+                assert host.run_test('rpm --rebuilddb'), 'Failed to rebuild the RPM db files'
+            except AssertionError:
+                workaround_error = 'chmod 755 /var/lock /var/lock/rpm ' \
```
narmaku, line 324: I suggest renaming this var to `change_permissions_cmd`.

test_suite/cloud/test_aws.py
```diff
@@ -317,8 +317,17 @@ def test_required_packages_are_installed(self, host):
     def test_rhui_pkg_is_installed(self, host):
         with host.sudo():
             # CLOUDX-590
-            assert host.run_test('rm -f /var/lib/rpm/__db.*'), 'Failed to remove the RPM db files'
-            assert host.run_test('rpm --rebuilddb'), 'Failed to rebuild the RPM db files'
+            try:
+                assert host.run_test('rm -f /var/lib/rpm/__db.*'), 'Failed to remove the RPM db files'
```
narmaku, line 321: We can also use host.run(cmd) and check for "failed" property (and remove try/catch):

```
if host.run('rm -f /var/lib/rpm/__db.*').failed or \
    host.run('rpm --rebuilddb').failed:
    ...
```
This way we can avoid having to catch the `AssertionError` exception and if other errors happen (improbable, but who knows), we will still be able to execute the workaround.

I also removed the assertion failure messages since we won't never see them if we use the workaround.

Another option is to always run the workaround, beforehand, but I am not sure about this.

narmaku (commented):
Looks good! Just a couple of comments.

test_suite/cloud/test_aws.py
sshmulev: ACK

test_suite/cloud/test_aws.py
sshmulev: ACK

narmaku (approved):
LGTM!

PR #290, opened 2024-01-18
AzureConfigBuilderV2: gracefully handle dot in release versions

thozza:
An osbuild-composer PR#3887 [1] introduces "dot-notation" for distro versions, meaning that a dot will be used to separate major and minor version in distro names. For backward compatibility, distro names without a dot will be still accepted for distros which used to accept it.

The CIV code parsing a VHD name was not expecting the version to contain any dot. This results in failures such as [2].

Modify the regular expression to gracefully handle potential dot in the version string.

[1] https://github.com/osbuild/osbuild-composer/pull/3887
[2] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5943494707#L3695

### DIFF
--- a/cloud/terraform/azure_config_builder_v2.py
+++ b/cloud/terraform/azure_config_builder_v2.py
@@ -112,9 +112,10 @@ class AzureConfigBuilderV2(BaseConfigBuilder):
             r"(?P<date>\d{4}\d{2}\d{2})\.\b(?:sp\.)?(?P<build_nr>\d+)\.(?P<arch>\S*)\.\bvhd"
         )
 
+        # thozza NB: the 'version' number may contain dot to separate the major and minor version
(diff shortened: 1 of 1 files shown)


jrusz:
It seems we need https://github.com/osbuild/cloud-image-val/pull/289 first to fix the CI

narmaku (approved):
LGTM! Thanks for the fix. 

We plan to remove this logic from CIV and add it where it makes more sense for our use cases (not the case for upstream CIV). It will be e.g. as part of a stage or wrapper script for CI in Jenkins/GitLab, etc. It will be easier to maintain too.

narmaku:
@thozza , new runners are merged into main. you may now rebase and push. CI should pass.

PR #291, opened 2024-01-21
CIV: Adding new TC for EFS Utils

sshmulev:
https://docs.google.com/document/d/1ENvebNiyJRqEcwp_64ybzt07Jtzjp0SI8F93yMEv2U4/edit?usp=sharing


test_suite/cloud/fixtures.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Use terraform script to initialize the env for the EFS Utils verification
+    """
+    # TODO
+    # Need to update the terraform script initialization to add also the NFS device for this test
+    function_instance = request.node.cls
+    # run terraform script
+
+    def finalizer():
+        function_instance.infra_controller.destroy_infra()
```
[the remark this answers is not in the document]
narmaku: Resolving conversation since we will not use a custom terraform script. We will deploy everything together with a new feature introduced here: https://github.com/osbuild/cloud-image-val/pull/293

test_suite/package/test_efs_utils.py
```diff
@@ -0,0 +1,37 @@
...
+            assert host.mount_point('/mnt/efs').exists
+
+            assert host.run_test(
+                f'dd if=/dev/zero of={test_file} bs=3K count=1'), 'Failed to write the file'
+
+            assert host.run_test(f'echo "{file_system_dns_name}:/ /mnt/efs efs _netdev,tls 0 0" >> "/etc/fstab"'), \
+                'Failed to write fstab'
+
+            checksum_before_reboot = host.file(test_file).md5sum
+            test_lib.reboot_host(host)
```
narmaku, line 36: Maybe you need to assign the new host retrieved by this lib function, and then do the checksum with that new host? 

```suggestion
            rebooted_host = test_lib.reboot_host(host)
            assert checksum_before_reboot == rebooted_host.file(test_file).md5sum
```
If it's working well with current implementation, that's fine. I remember we had issues with rebooting the instances and that's why we decided to retrieved a new host built from the lib function.

test_suite/package/test_efs_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        Check basic functionality of EFS utils:
+        - Create a mount point and mount the EFS file system to the RHEL instance.
+        - Create a file on the mount point.
+        - Automate mount on boot.
+        - Checksum the file before reboot.
+        - reboot and checksum that the file is correct.
+        """
+        # will get 'file_system_dns_name' arg form the fixture that gets the output of the Terraform script
+        file_system_dns_name = instance_data['efs_file_system_dns_name']
+        test_file = '/mnt/efs/testfile'
```
narmaku, line 21: I would be cautious with the file name. It can have issue when running on more than 1 instance belonging to the same region, they will share the efs file system and if they both try to write on the same file, it can have unexpected results.

I propose to append a random string at the end of the filename, or to the whole filename.

E.g.:
```
import uuid

...

test_file = '/mnt/efs/{}'.format(str(uuid.uuid4()))
```

test_suite/package/test_efs_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        with host.sudo():
+            assert host.run_test(
+                f'mkdir /mnt/efs && mount -t efs {file_system_dns_name} /mnt/efs/'
+            ), f'Failed mount {file_system_dns_name} to efs folder'
+
+            assert host.mount_point('/mnt/efs').exists
```
narmaku, line 27: :clap: I loved this!

test_suite/package/test_efs_utils.py
narmaku, line 1: We are using `/mnt/efs/` mount point string in multiple parts of the code. To ease maintenance and follow best practices I would recommend to create a variable for this, then refer to the var inside the command strings and wherever else needed.

test_suite/package/test_efs_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        - Checksum the file before reboot.
+        - reboot and checksum that the file is correct.
+        """
+        # will get 'file_system_dns_name' arg form the fixture that gets the output of the Terraform script
+        file_system_dns_name = instance_data['efs_file_system_dns_name']
+        test_file = '/mnt/efs/testfile'
+        with host.sudo():
+            assert host.run_test(
+                f'mkdir /mnt/efs && mount -t efs {file_system_dns_name} /mnt/efs/'
+            ), f'Failed mount {file_system_dns_name} to efs folder'
```
narmaku, line 25:
```suggestion
            ), f'Failed mount {file_system_dns_name} to efs directory'
```
Linux naming convention is better for us ;)

test_suite/package/test_efs_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+            assert host.mount_point('/mnt/efs').exists
+
+            assert host.run_test(
+                f'dd if=/dev/zero of={test_file} bs=3K count=1'), 'Failed to write the file'
+
+            assert host.run_test(f'echo "{file_system_dns_name}:/ /mnt/efs efs _netdev,tls 0 0" >> "/etc/fstab"'), \
+                'Failed to write fstab'
+
+            checksum_before_reboot = host.file(test_file).md5sum
```
narmaku, line 35: I love this! Excellent use of the testinfra modules capabilities! :clap:

narmaku (changes requested):

narmaku:
Reminder: We need to wait for #293 and #300 to be merged before we test (locally) with the new efs deployment feature and specifying custom test suite file path.

test_suite/package/test_efs_utils.py
narmaku: Tested and it does not affect the code at all.

test_suite/package/test_efs_utils.py
narmaku: Added EPOCH timestamp instead.

PR #292, opened 2024-01-24
Enable boot diagnostics for Azure VMs provisioned by the tool.

narmaku:
Thi will help us debug boot issues not detected by the Activity Log or terraform.

### DIFF
--- a/cloud/terraform/base_config_builder.py
+++ b/cloud/terraform/base_config_builder.py
@@ -38,9 +38,23 @@ class BaseConfigBuilder:
         :param separator: String to be used as a separator of the items passed in resource_names_combination
         :return: String composed of a prefix, resource names combination and a suffix with random numbers.
         """
-        combinations = [self.resource_name_prefix]
-        combinations.extend(resource_names_combination)
-        combinations.append(self.get_random_numbers())
+        tf_resource_name_max_chars = 63
+        random_num = self.get_random_numbers()
+
+        # If any of the names already have prefixes, those will be removed to avoid redundancy.
+        # Example: We received ['civ-network', 'vnc'] as resource_names_combination argument.
+        # Considering self.resource_name_prefix is "civ" and separator "-", combined_name will contain: "network-vnc"
+        combined_name = separator.join(resource_names_combination).replace(
+            f'{self.resource_name_prefix}{separator}', '')
+
+        # We calculate the full length after concatenation with prefix, two separators and random numbers
+        end_index = tf_resource_name_max_chars - len(self.resource_name_prefix) - len(random_num) - (len(separator) * 2)
+
+        combinations = [
+            self.resource_name_prefix,
+            combined_name[0:end_index],
+            random_num
+        ]
 
         return separator.join(combinations)
 
--- a/cloud/terraform/azure_config_builder_v2.py
+++ b/cloud/terraform/azure_config_builder_v2.py
@@ -315,6 +315,10 @@ class AzureConfigBuilderV2(BaseConfigBuilder):
             'public_key': f'${{file("{self.ssh_key_path}")}}'
         }
 
+        boot_diagnostics = {
+            'storage_account_uri': None
+        }
+
         new_instance = {
             'name': instance_hostname,
             'location': instance['location'],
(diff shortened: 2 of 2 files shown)


cloud/terraform/base_config_builder.py
```diff
@@ -38,9 +38,21 @@ def create_resource_name(self, resource_names_combination, separator='-') -> str
...
+            chars_to_strip = combined_name_full_length - tf_resource_name_max_chars
+            combined_name = combined_name[0:-chars_to_strip]
+
+        combinations = [
+            self.resource_name_prefix,
+            combined_name,
+            random_num
+        ]
 
         return separator.join(combinations)
```
atodorov, line 57: Why not use a temporary variable and move the name truncation inside the `return` statement ? 

If you mean to keep the prefix but still struncate the rest of the name then why not construct the part after the prefix and truncate it before return ?

cloud/terraform/base_config_builder.py
```diff
@@ -38,9 +38,21 @@ def create_resource_name(self, resource_names_combination, separator='-') -> str
...
-        combinations.append(self.get_random_numbers())
+        tf_resource_name_max_chars = 63
+        random_num = self.get_random_numbers()
+
+        combined_name = separator.join(resource_names_combination).replace(f'{self.resource_name_prefix}-', '')
+        combined_name_full_length = len(separator.join([self.resource_name_prefix, combined_name, random_num]))
+
+        if combined_name_full_length > tf_resource_name_max_chars:
+            chars_to_strip = combined_name_full_length - tf_resource_name_max_chars
+            combined_name = combined_name[0:-chars_to_strip]
```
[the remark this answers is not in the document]
narmaku: Good point, no need to check the length since `[0:63]` will either provide the scripted string, or the full length if less than 63 chars!

cloud/terraform/base_config_builder.py
atodorov: This question is still valid. From what I can tell the entire function is almost equivalent to 

```
        combinations = [self.resource_name_prefix]
        combinations.extend(resource_names_combination)
        combinations.append(self.get_random_numbers())

        return separator.join(combinations)[0:63]
```

Given the existing doc-string
```
        :param resource_names_combination: List of keywords, names or IDs that you want to include in the resource name.
                                           Example: ['westeurope', 'vm']
```

I don't think that the middle bits are the ones which are too long so it doesn't really make any difference if characters are removed from the middle or from the end. 

I'm not quite sure what you are trying to do by changing this function that much, hoever it feels to me that in practice it only needs a very small change - the string slice operation at the end of the `return` statement.  Adding ~ 5 chars instead of unnecessarily complicating the logic of this function seems like a better option to me.

atodorov (changes requested):

cloud/terraform/base_config_builder.py
narmaku: Hello @atodorov , let me explain:
Sometimes we pass long names here, and if we don't truncate the "middle" part of it, it may end up with a duplicated resource name for terraform and making the provisioning fail. I encountered this issue by myself by running things locally and the name can sometimes be long enough. If we just truncate after the concatenation, the random numbers at the end that provide the unique identifier as a resource name,, may be truncated as well and that's when we encounter the "duplicated resource" issue.

In the end, we should a void super long names because they are not very user-friendly for debugging, however some names are taken from the image metadata itself, such as the vhd blob filename or what's defined in resources.json as image name, etc. We cannot control that.

cloud/terraform/base_config_builder.py
```diff
@@ -38,9 +38,18 @@ def create_resource_name(self, resource_names_combination, separator='-') -> str
         :param separator: String to be used as a separator of the items passed in resource_names_combination
         :return: String composed of a prefix, resource names combination and a suffix with random numbers.
         """
-        combinations = [self.resource_name_prefix]
-        combinations.extend(resource_names_combination)
-        combinations.append(self.get_random_numbers())
+        tf_resource_name_max_chars = 63
+        random_num = self.get_random_numbers()
+
+        combined_name = separator.join(resource_names_combination).replace(f'{self.resource_name_prefix}-', '')
```
atodorov, line 44: Is the prefix ever part of `resource_names_combination` ?  Removing it feels a bit redundant if it isn't part of the input:

cloud/terraform/base_config_builder.py
atodorov: Thanks for explanation. This now makes more sense.

atodorov (commented):
Added a question

cloud/terraform/base_config_builder.py
narmaku: In the cloud-specific config builders we use some dependent resource names and they do have the prefix already. This is to avoid having the prefix set multiple times, which also increases the length and adds extra prefixes where it shouldn't. That's why we remove it in case it was already added previously.

atodorov: :man_facepalming: 

This really needs to be documented in the source code because it isn't obvious why it needs to work like that.

atodorov (approved):

PR #293, opened 2024-01-30
Added efs creation and its dependent resources in new class AWSConfigBuilderEfs

narmaku:
The resources include: 
* aws_efs_file_system
* aws_efs_mount_target

The data sources include:
* aws_vpc (for each region)
* aws_subnets (list of subnets available on each VPC, per region)

A new class has been introduced: AWSConfigBuilderEfs. It is in charge of deploying the EFS file system and mount targets along with the AMI instances.

Instance metadata (from instance_data fixture) will contain information about the efs file system's DNS name, so it can be used within the test cases.

For the new class to be used and deploy efs resources, we need to pass "aws-efs" as one of the tags in CIV params/config.

### DIFF
--- a/cloud/terraform/aws_config_builder_efs.py
+++ b/cloud/terraform/aws_config_builder_efs.py
@@ -1,0 +1,191 @@
+from cloud.terraform.base_config_builder import BaseConfigBuilder
+
+
+class AWSConfigBuilderEfs(BaseConfigBuilder):
+    cloud_name = 'aws'
+    cloud_provider_definition = {'aws': {'source': 'hashicorp/aws', 'version': '~> 4.62.0'}}
+
+    def build_providers(self):
+        all_regions = self.__get_all_regions_from_resources_file()
+        for region in all_regions:
+            self.providers_tf['provider'][self.cloud_providers[self.cloud_name]] \
+                .append(self.__new_aws_provider(region))
+
+        return self.providers_tf
+
+    def __get_all_regions_from_resources_file(self):
+        instances_regions = [i['region'] for i in self.resources_dict['instances']]
+
+        return list(dict.fromkeys(instances_regions))
+
+    def __new_aws_provider(self, region):
+        return {
+            'region': region,
+            'alias': region,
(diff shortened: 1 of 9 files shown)


sshmulev (approved):
LGTM! +1

narmaku:
Declaring additional providers per region could potentially solve this issue, as documented here:
* https://discuss.hashicorp.com/t/get-vpc-id-which-is-in-different-region/21936/2

cloud/terraform/terraform_controller.py
```diff
@@ -59,9 +55,29 @@ def get_instances(self):
...
     def get_instances_aws(self, resources):
-        instances_info = {}
+        regional_efs_file_systems = {}
+        for resource in resources:
+            if resource['type'] == 'aws_efs_file_system':
+                efs_dns_name = resource['values']['dns_name']
+                result = re.match(r'fs-.*\.efs\.(.*).amazon', efs_dns_name)
+
+                if not result:
+                    raise Exception(f'Could not et EFS file system region in DNS name: {efs_dns_name}')
```
narmaku, line 74:
```suggestion
                    raise Exception(f'Could not get EFS file system region in DNS name: {efs_dns_name}')
```

cloud/terraform/aws_config_builder_efs.py
```diff
@@ -0,0 +1,191 @@
...
+        declared_aws_subnets = 'data.aws_subnets.{}.ids'.format(instance['aws_subnets'])
+        declared_filesystem_id = 'aws_efs_file_system.{}.id'.format(instance['aws_efs_file_system'])
+
+        new_efs_mount_target = {
+            'provider': f'aws.{region}',
+            'for_each': f'${{toset({declared_aws_subnets})}}',
+            'file_system_id': f'${{{declared_filesystem_id}}}',
+            'subnet_id': '${each.value}',
+        }
+
```
narmaku, line 155: Add tags: `self.add_tags(self.config, new_efs_mount_target)`

PR #294, opened 2024-02-06
Use correct detection of modified cfg in test_cloud_cfg()

plageat:
Hi,

It seems, that generic check **test_cloud_cfg()** always **PASS**, because **_rpm -V_** only returns **_0_** or **_1_** which makes **host.run_test()** always succeed.

The proposal is to check modification flags of **/etc/cloud/cloud.cfg** in a more accurate way:

`rpm -Vf /etc/cloud/cloud.cfg | grep -e "^S.5.*/etc/cloud/cloud.cfg$"`

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -585,11 +585,12 @@ class TestsCloudInit:
         """
         Verify file /etc/cloud/cloud.cfg is not changed
         """
-        package_to_check = 'cloud-init'
+        cloud_cfg = '/etc/cloud/cloud.cfg'
+        verify_cmd = f'rpm -Vf {cloud_cfg} | grep -e "^S.5.*{cloud_cfg}$"'
 
         with host.sudo():
-            assert host.run_test(f'rpm -V {package_to_check}'), \
-                f'There should not be changes in {package_to_check} package'
+            assert not host.run(verify_cmd).stdout, \
+                f'There should not be changes in {cloud_cfg} package'
 
     @pytest.mark.run_on(['rhel9.0'])
     def test_cloud_cfg_netdev_rhel9(self, host):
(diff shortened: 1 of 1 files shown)


test_suite/generic/test_generic.py
```diff
@@ -585,11 +585,12 @@ def test_cloud_cfg(self, host):
         """
         Verify file /etc/cloud/cloud.cfg is not changed
         """
-        package_to_check = 'cloud-init'
+        cloud_cfg = '/etc/cloud/cloud.cfg'
+        verify_cmd = f'rpm -Vf {cloud_cfg} | grep -e "^S.5.*{cloud_cfg}$"'
```
thozza, line 589: I think that you should consider also files shipped by the package in /etc/cloud/cloud.cfg.d`

test_suite/generic/test_generic.py
```diff
@@ -585,11 +585,12 @@ def test_cloud_cfg(self, host):
...
 
         with host.sudo():
-            assert host.run_test(f'rpm -V {package_to_check}'), \
-                f'There should not be changes in {package_to_check} package'
+            assert not host.run(verify_cmd).stdout, \
+                f'There should not be changes in {cloud_cfg} package'
```
thozza, line 593: The comment does not make sense any more. `cloud_cfg` is not a package AFAICT.

test_suite/generic/test_generic.py
plageat: Sounds like writing completely different test
Of course, no problem for me

test_suite/generic/test_generic.py
plageat: Fixed in 16aedb1cc4427de8879b287ed0db459c9af76639

test_suite/generic/test_generic.py
plageat: Fixed in 16aedb1cc4427de8879b287ed0db459c9af76639

thozza: > Sounds like writing completely different test Of course, no problem for me

True, but the test originally covered also `/etc/cloud/cloud.cfg.d`, even though it was probably not the intention based on the doc string.

thozza (approved):
LGTM

PR #296, opened 2024-03-08
Skipped test_cloud_configs test case in RHEL-SAP 8.4 and older.

narmaku:
Related JIRA: CLOUDX-812. These images were not built by image builder and do not contain the changes we check in the test case.


sshmulev (approved):
+1

PR #298, opened 2024-03-12
Skipped test_no_avc_denials test case due to a known issue.

narmaku:
RHEL-24346: NetworkManager causing avc denials, to be fixed on RHEL-9.5.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -16,6 +16,11 @@ class TestsGeneric:
         """
         Check there is no avc denials (selinux).
         """
+        if host.system_info.distribution == 'rhel' and \
+                float(host.system_info.release) == 9.4:
+            pytest.skip('RHEL-24346: Skipping on RHEL-9.4 due to a known issue with NetworkManager.')
+
+        # CLOUDX-320: This "if" is pending to be removed
         if instance_data['cloud'] == 'azure':
             pytest.skip('Skipping on fedora due to old image definitions')
 


narmaku:
I decided to be mores specific in the check before skipping the test case.
This TC is quite important in our test suite and should only be skipped IF the system is redhat-9.4.

narmaku:
Nightly pipeline manually run [here](https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/pipelines/1209775659) targeting this PR code.

test_suite/generic/test_generic.py
```diff
@@ -16,6 +16,11 @@ def test_no_avc_denials(self, host, instance_data):
         """
         Check there is no avc denials (selinux).
         """
+        if host.system_info.distribution == 'rhel' and \
+                float(host.system_info.release) == 9.4:
+            pytest.skip('RHEL-24346: Skipping on RHEL-9.4 due to a known issue with NetworkManager.')
```
atodorov, line 21: Q: is there a corresponding ticket to re-enable this check once selinux-policy is fixed ? Ideally we'd like to have this execute in future versions of 9.4 instead of being skipped for the entire duration of he minor release.

narmaku: +1. I created this ticket for future reference. We will see it each time we review backlog items (at least once per Q).
https://issues.redhat.com/browse/CLOUDX-848

thozza (approved):

atodorov (approved):

PR #300, opened 2024-03-19
Added new param/config to specify custom test suite paths.

narmaku:
It is an optional argument/config value. If not used, behavior will remain the same. 

This means that we still hardcode the default test suite paths which belong to test_suite/cloud and test_suite/generic directories.

### DIFF
--- a/test_suite/suite_runner.py
+++ b/test_suite/suite_runner.py
@@ -32,25 +32,37 @@ class SuiteRunner:
         self.debug = debug
 
     def run_tests(self,
+                  test_suite_paths,
                   output_filepath,
                   test_filter=None,
                   include_markers=None):
         if os.path.exists(output_filepath):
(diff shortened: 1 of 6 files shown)


test_suite/suite_runner.py
```diff
@@ -79,9 +91,9 @@ def compose_testinfra_command(self,
 
         return ' '.join(command_with_args)
 
-    def get_test_suite_paths(self):
+    def get_default_test_suite_paths(self):
         """
-        :return: A String array of test file absolute paths that will be executed in the cloud instances
+        :return: A list of ste suite file paths that will be used in case there are no test suites passed as argument
```
narmaku, line 96:
```suggestion
        :return: A list of test suite file paths that will be used in case there are no test suites passed as argument
```

sshmulev (approved):
+1

PR #306, opened 2024-04-04
Improved "Preparing environment" step in main script flow.

narmaku:
The section was lacking console output, so end-user did not know what was going on during this step. Furthermore, if something failed, there was no visibility on what was exactly the failure.

The changes include:
- Add a new prepare_environment() function to become more modular and extend it over time
- Improve console output by informing about the tasks that are being performed during the preparation of the environment
- Refactor and parallelize the copy of team's public ssh keys (extra), hopefully saving more time when deploying ~180 images simultaneously
- Move hardcoded tmp file paths to default configuration, so they can be customized and referenced from other parts of the code
- Add a new method to the CIVConfig class, to export CIV config items to env variables, so they can be easily referenced from other parts of the code


narmaku:
You can see an example of the console output here:
* https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/6578277145#L1925

atodorov (commented):
This should be reviewed by the rest of Cloud X team. I cannot really review in depth b/c I'm not familiar with the cloud-image-val source code that much or how it is supposed to be changing in the future and it is outside of my domain.

PR #307, opened 2024-04-10
Implemented packages testing automation

narmaku:
We introduce a new job in .gitlab-ci.yml. It is about enabling the CloudX team to run it on-demand, and install packages in AWS AMIs to be built by the CI. This will unblock us in many ways, and we will be able to automatically test new packages we plan to add to RHEL from now on.

This is not a final solution, since we must build our own CI infrastructure as a definitive solution.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -106,7 +106,7 @@ prepare-rhel-internal-runners:
 .rhel_runners: &rhel_runners
   RUNNER:
     - aws/centos-stream-8-x86_64
-    - aws/rhel-8.4-ga-x86_64
+    # - aws/rhel-8.4-ga-x86_64
     - aws/rhel-8.9-ga-x86_64
     - aws/rhel-9.3-ga-x86_64
     - aws/rhel-8.10-nightly-x86_64
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,7 +6,7 @@ stages:
 
 .deps:
   before_script:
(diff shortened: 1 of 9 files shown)


narmaku:
To run on-demand, you can simply modify some of the variables from this scheduled pipeline:
* https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/pipeline_schedules/1549645/edit?id=1549645
Please note that the pipeline itself is disabled, so it can only be run on-demand.

ConorC117 (commented):
Code looks good to me! I watched the walk-through as well and found it very clear :) My only question is about the Schutzbot check and whether this is expected to fail at the moment?

narmaku:
> Code looks good to me! I watched the walk-through as well and found it very clear :) My only question is about the Schutzbot check and whether this is expected to fail at the moment?

Thanks @ConorC117 !
About rhel-8.4 failure in the CI:  I just confirmed with the Image Builder folks and that runner is broken, we should skip it for now, until it is solved in the future. They did exactly the same in their CI two weeks ago.

FKolwa (approved):
>About rhel-8.4 failure in the CI: I just confirmed with the Image Builder folks and that runner is broken, we should skip it for now, until it is solved in the future. They did exactly the same in their CI two weeks ago.

I was about to comment on that. Thanks for clarifying. In that case I'd give you a go from my side.
Implementation looks good for now. I'm 100% behind you on getting our own CI infra but for the time being this is a great solution. Thanks for doing this.

PR #311, opened 2024-04-16
test_network_manager_cloud_setup: allow alternate quoting

### LINKED ISSUE #310: Failing tests due to string quoting changes in systemd unit files
In https://github.com/osbuild/images/pull/567 we made a non-functional change to the way we write systemd unit files.  The `Environment` option in the osbuild stage now supports multiple entries.  At the same time, we made it so that `key=value` assignments are quoted, which makes it easier to work with values with spaces.
From the systemd-exec man page:
```
ENVIRONMENT
       Environment=
           Sets environment variables for executed processes. Each line is unquoted using the rules described in "Quoting" section in
           systemd.syntax(7) and becomes a list of variable assignments. If you need to assign a value containing spaces or the equals sign to a
           variable, put quotes around the whole assignment.
...

           Example:

               Environment="VAR1=word1 word2" VAR2=word3 "VAR3=$word 5 6"
```

I think this change made CIV start failing to validate because it expects the unquoted strings in some of the cloud systemd units like here:
https://github.com/osbuild/cloud-image-val/blob/32b5842a751d10ffe8951f82b04ca5240b700534/test_suite/generic/test_generic.py#L716

croissanne:
Newer versions of osbuild & images quote environment variables in systemd unit files.

Fixes #310

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -709,11 +709,17 @@ class TestsNetworking:
         files_and_configs_by_cloud = {
             'aws': {
                 'file_to_check': os.path.join(cloud_setup_base_path, '10-rh-enable-for-ec2.conf'),
-                'expect_config': 'Environment=NM_CLOUD_SETUP_EC2=yes'
+                'expect_configs': [
+                    'Environment=NM_CLOUD_SETUP_EC2=yes',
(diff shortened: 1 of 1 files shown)


achilleas-k (commented):
THANK YOU!!

achilleas-k (approved):
THANK YOU!!

achilleas-k (approved):

PR #312, opened 2024-04-16
test_no_avc_denials: skip on centos 9 due to RHEL-24346

croissanne:


bcl (approved):
Looks ok for now, but at some point I think all the floats in this codebase need to be replaced with version.parse() comparisons.

croissanne:
Admin merging, centos-9 was skipped, the other ones weren't.

PR #316, opened 2024-04-25
CIV: Add new test for otel package

sshmulev:


test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+import logging
+import time
+import json
+
+from test_suite.generic.test_generic import TestsSubscriptionManager
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture()
+def check_instance_status(request, instance_data, host):
```
narmaku, line 11: Why do we need this fixture?
The instance will already be operative and ssh-ready at the moment we run this test suite.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.fixture()
+def install_packages(request, host):
+    class_instance = request.node.cls
+    install_cmd = f'yum install -y  {class_instance.package_name} --nogpgcheck'
+    repo_path = ('https://copr.fedorainfracloud.org/coprs/miyunari'
+                 '/redhat-opentelemetry-collector/repo/rhel-9/miyunari-redhat-opentelemetry-collector-rhel-9.repo')
+    with host.sudo():
+        assert host.run_test(f'yum-config-manager --add-repo {repo_path}')
```
narmaku, line 54: As we discussed the other day, we should not use `host.run_test(cmd)` anymore, since it also returns `True` when the command failed and exited with 1. Instead, please use `host.run(cmd).succeeded`.

narmaku: Same for the rest of the code.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+        assert host.run_test(install_cmd), f'Failed to install the package {class_instance.package_name}'
+        output = host.check_output("rpm -qa | grep opentelemetry*")
```
narmaku, line 56: Please use `pytest-testinfra` modules whenever possible, to avoid having to manually run commands for things that our framework already do for you.
Please check [host.package](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.package.Package)

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+        print(output)
+
+
+@pytest.fixture()
+def start_service(request, host):
+    class_instance = request.node.cls
+    start_enable_service = (f'systemctl start {class_instance.package_name} && '
```
narmaku, line 63: Please use pytest-testinfra modules whenever possible, to avoid having to manually run commands for things that our framework already do for you.
Please check [host.service](https://testinfra.readthedocs.io/en/latest/modules.html#service)

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+                            f'systemctl enable {class_instance.package_name}')
+    is_active = (f'systemctl is-active {class_instance.package_name}')
+
+    with host.sudo():
+        assert host.run(start_enable_service).succeeded, (f'Failed to start the service {class_instance.package_name}')
+        assert host.run(is_active).succeeded, (f'Failed to activate the service {class_instance.package_name}')
+
+    def finalizer():
+        logger.info(f'Removing the package {class_instance.package_name}')
```
narmaku, line 72: I like this approach, however I think we need to introduce a logging mechanism in our ticket dedicated for that and make it standard across our project and test suites.

So, for now let's try to be consistent with the current implementations and use our own lib functions.

I suggest to use `test_lib.print_host_command_output()` to output the command and its result.
I also suggest to use `console_lib.print_divider()` to print a "section" in your test case outputs.
you can then simply call `print()` function to print more information inside each section.

See how it has been implemented in this test case as an example: https://github.com/osbuild/cloud-image-val/blob/144f69c99a28c173911fa59abc4890ca06df1b94/test_suite/package/test_efs_utils.py#L33

```suggestion
        console_lib.print_divider(f'Removing the package {class_instance.package_name}')
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+            - Make a failure ssh connection to the instance.
+            - Check for error messages in the ssh logs cotaining within the instance.
+            - Check the error logs with AWS CLI and compare it to the logs in "/var/log/secure".
+        Finalize:
+            - Remove the package from the instance and verify it's not present anymore.
+            - Try a failure ssh again and check that the logs don't appear.
+        """
+        logger.info("Connect to the instance without a key in order to fail")
+        instance_dns = instance_data['public_dns']
+        assert not host.backend.run_local(f'ssh {instance_dns}')
```
narmaku, line 126:
```suggestion
        assert not host.backend.run_local(f'ssh {instance_dns}').succeeded
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture(scope="function", autouse=True)
+def run_subscription_manager_auto(request, host, instance_data):
+    class_instance = request.node.cls
+    # Run the subscription manager auto test before any tests in this file
+    TestsSubscriptionManager.test_subscription_manager_auto(class_instance, host, instance_data)
+
+
+@pytest.mark.package
+@pytest.mark.run_on(['rhel9.1'])
```
narmaku, line 89: I guess this will change when we do final testing, to something like `@pytest.mark.run_on(['>=rhel9.5'])`.
Now, maybe we should limit it to rhel9.4+?
```suggestion
@pytest.mark.run_on(['>=rhel9.4'])
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+    def check_aws_cli_logs(self, host):
+        command_to_run = [
+            'aws', 'logs', 'filter-log-events',
+            '--log-stream-names', 'testing-integrations-stream-emf',
+            '--filter-pattern', 'Invalid',
+            '--log-group-name', 'testing-logs-emf'
+        ]
+        run_aws_cli_cmd = ' '.join(command_to_run)
+
+        command_output = host.backend.run_local(run_aws_cli_cmd), 'Failed to get log output'
```
narmaku, line 102:
```suggestion
        command_output = host.backend.run_local(run_aws_cli_cmd).stdout
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+        assert "Invalid" in command_output
+
+    @pytest.mark.usefixtures(
+        check_instance_status.__name__,
+        install_packages.__name__,
+        modify_iam_role.__name__,
+        start_service.__name__,
+    )
+    def test_otel(self, host, instance_data):
```
narmaku, line 111: I think this test case is overloaded with too many fixtures that could potentially be other test cases or change the running conditions:
* `run_subscription_manager_auto` should not be run every time a test case is run. This is only needed by `install_packages`.
* `install_packages` and  `modify_iam_role` could be fixtures to be run once, at a class level.
* `start_service` should be a test case that runs first. E.g. `test_otel_service`.
  * the `finalizer` should be another test case, that runs in the end. E.g. `test_no_logs_after_removing_otel`.

narmaku: To change the order in which the test cases are run, please use the `order(num)` marker.
You can find examples in other test cases from the generic or cloud-specific test suites.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture()
+def check_instance_status(request, instance_data, host):
+    instance_id = instance_data['instance_id']
+    command_to_run = [
+        'aws', 'ec2', 'describe-instance-status',
+        '--instance-ids', instance_id
+    ]
+
+    while True:
```
narmaku, line 18: Add a safety net/timeout to this while, it can enter into an infinite loop.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+    ]
+    modify_iam_role_cmd = ' '.join(command_to_run)
+
+    assert host.backend.run_local(modify_iam_role_cmd), 'faild to update iam role'
+
+
+@pytest.fixture()
+def install_packages(request, host):
+    class_instance = request.node.cls
+    install_cmd = f'yum install -y  {class_instance.package_name} --nogpgcheck'
```
narmaku, line 50:
```suggestion
    install_cmd = f'yum install -y {class_instance.package_name} --nogpgcheck'
```

test_suite/package/test_otel.py
narmaku: Maybe in this case it is because you want to list all of the matching packages, in that case I think it's perfect. I would just suggest to use `test_lib.print_host_command_output()`. It will automatically print all the packages for your so you don't have to store cmd output in a var to later print it out.

```suggestion
        test_lib.print_host_command_output("rpm -qa | grep opentelemetry*")
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+@pytest.fixture()
+def start_service(request, host):
+    class_instance = request.node.cls
+    start_enable_service = (f'systemctl start {class_instance.package_name} && '
+                            f'systemctl enable {class_instance.package_name}')
+    is_active = (f'systemctl is-active {class_instance.package_name}')
+
+    with host.sudo():
+        assert host.run(start_enable_service).succeeded, (f'Failed to start the service {class_instance.package_name}')
+        assert host.run(is_active).succeeded, (f'Failed to activate the service {class_instance.package_name}')
```
narmaku, line 69: As commented before, check if you can use the Package module from testinfra.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @pytest.mark.usefixtures(
+        check_instance_status.__name__,
+        install_packages.__name__,
+        modify_iam_role.__name__,
+        start_service.__name__,
+    )
+    def test_otel(self, host, instance_data):
+        """
+        Verify basic funstionality for Otel package:
```
narmaku, line 113:
```suggestion
        Verify basic functionality for OpenTelemetry (OTEL) package:
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+            - Check the error logs with AWS CLI and compare it to the logs in "/var/log/secure".
+        Finalize:
+            - Remove the package from the instance and verify it's not present anymore.
+            - Try a failure ssh again and check that the logs don't appear.
+        """
+        logger.info("Connect to the instance without a key in order to fail")
+        instance_dns = instance_data['public_dns']
+        assert not host.backend.run_local(f'ssh {instance_dns}')
+        logger.info("Check for error logs in the instance logs")
+        host_output = host.check_output('sudo tail -F /var/log/secure')
```
narmaku, line 128: If it's about checking for file contents, you can use the testinfra [File](https://testinfra.readthedocs.io/en/latest/modules.html#file) module.

narmaku (changes requested):

test_suite/package/test_otel.py
sshmulev: In order to make changes in the security of the instance it is required that the instance status and system_status will be passed/ok status, otherwise it will not be possible to make this change.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture()
+def modify_iam_role(request, instance_data, host):
+    instance_id = instance_data['instance_id']
+    iam_role_name = "CloudWatchAgentServerRole_2"
+
+    command_to_run = [
+        'aws', 'ec2', 'associate-iam-instance-profile',
+        '--instance-id', instance_id,
+        '--region', 'us-west-1',
```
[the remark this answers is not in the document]
sshmulev: Yes I agree, but since we can't generate our own AMIs for upstream versions yet, there are pretty much workarounds here, this one is defenitly a thing we should correct when generating at least nighly build will be available for us.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,132 @@
...
+        system_status = cmd_output["InstanceStatuses"][0]["SystemStatus"]["Status"]
+
+        if instance_status == "ok" and system_status == "ok":
+            break
+        logger.info("Instance status is not 'passed' yet. Waiting...")
+        time.sleep(5)
+
+
+@pytest.fixture()
+def modify_iam_role(request, instance_data, host):
```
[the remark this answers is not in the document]
sshmulev: ACK

test_suite/package/test_otel.py
sshmulev: ACK

test_suite/package/test_otel.py
sshmulev: regarding the other I couldn't find in test infra commands that support um install or config manager, so I will leave it as it is (line 54,55)
If you find something, so please update me!:)

test_suite/package/test_otel.py
sshmulev: I couldn't find some module that will start and activate the service but I changed the verification for that

narmaku: Got it. It seems that the service module is useful for verification but not for making changes! Thanks for double-checking.

test_suite/package/test_otel.py
narmaku: Yeah, I think there's nothing related to adding repos and installing packages, only querying them. In this case it makes total sense to query all packages to then get a list with grep, that maches all the OTEL-related pkgs.

The package module can query only one specific package with a specific name, so that would mean we need to create a list of otel pkg names to check and iterate over them. Not a fan of that.

test_suite/package/test_otel.py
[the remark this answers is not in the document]
sshmulev: Yes I will change this part

test_suite/package/test_otel.py
sshmulev: ACK

test_suite/package/test_otel.py
sshmulev: I couldn't find some module that will start and activate the service but I changed the verification for that

sshmulev:
Added the changes will push them

test_suite/package/test_otel.py
narmaku: Got it! I thought that the status of the instance was already ok after CIV deployed it.
Out of curiosity: Do we know how much time does it take for the instance to be "ready" for security changes, after we deploy it?

sshmulev: mmm it takes about 2-3 minutes

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,149 @@
+import pytest
+import logging
+import time
+import json
+from lib import console_lib
+from test_suite.generic.test_generic import TestsSubscriptionManager
+from lib import test_lib
+logger = logging.getLogger(__name__)
```
narmaku, line 8: Is there a reason why we are keeping this logger after the last review conversation?

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        assert host.service(class_instance.package_name).is_running, (
+            f'Failed to run the service {class_instance.package_name}'
+        )
+
+    def finalizer():
+        console_lib.print_divider(f'Removing the package {class_instance.package_name}')
+        assert host.run(f'sudo yum remove -y {class_instance.package_name}')
+        assert not host.check_output(f'rpm -q {class_instance.package_name}')
+        assert not host.run(f'ssh {class_instance.instance_dns}').succeeded
```
narmaku, line 88:
```suggestion
        assert host.run(f'ssh {class_instance.instance_dns}').failed
```

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,149 @@
...
+        logger.info("Verify logs don't appear")
```
narmaku, line 89: As discussed in last review, please use `print()` or current CIV libs for printing test case console outputs.
Same for the rest of the code.

test_suite/package/test_otel.py
```diff
@@ -0,0 +1,149 @@
...
+        # TODO: verify logs don't appear.
+    request.addfinalizer(finalizer)
+
+
+@pytest.fixture(scope="function", autouse=True)
```
narmaku, line 94: I am still iffy about autousing this fixture at the function level. Auto-registration is time-consuming.

Today this file has only one test case. However, _if by any chance_ we add more test cases and forget about this, it will take too much time to run, and it may also fail since the system will already be subscribed. Let's make it run only once.

narmaku (changes requested):

test_suite/package/test_otel.py
sshmulev: No problem. changing the scope to "session", this ensures that the fixture is executed only once, regardless of how many test functions or classes are using it throughout the entire test session.

narmaku:
@sshmulev did the tests passed in your local environment? If so, let's merge this! :rocket:

narmaku (approved):
LGTM - just make sure tests pass 100% in your local environment (since we cannot test this any other way right now).

test_suite/package/test_otel.py
```diff
@@ -52,14 +52,33 @@ def modify_iam_role(instance_data, host):
     assert host.backend.run_local(modify_iam_role_cmd), 'faild to update iam role'
```
narmaku, line 52: `'Failed to update IAM role'`

narmaku (approved):
LGTM, just a small comment.

narmaku (changes requested):

test_suite/generic/helpers.py
```diff
@@ -0,0 +1,10 @@
+from test_suite import conftest
+
+
+def instance_data(host):
```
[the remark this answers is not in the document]
sshmulev: I mentioned in this comment that since this fixture is limited by a class scope of "function", it is impossible to use it in other fixtures like in this case which are using class scope for example.

There are two solutions for this:
1. Change the existing fixture of instance_data to class scope
2. Create an independent function (like my solution in helpers.py) that can be used in any fixture or method and it is not limited by any class scope.

test_suite/package/otel_package/fixtures.py
```diff
@@ -0,0 +1,111 @@
+import json
+import time
+import pytest
+from lib import console_lib
+from lib import test_lib
+from test_suite.generic import helpers
+
+
+@pytest.fixture(scope='class')
+def initialize_veribales(request, host):
```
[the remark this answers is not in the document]
sshmulev: Ack, Thanks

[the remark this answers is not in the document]
sshmulev: "you were having issue when using fixtures inside another fixture." - This is exactlly the issue I replied on your comment above (pytest doen't support using fixtures of function scope inside a class scope fixture for example) this is why I added this as a function rather using this "instance_data" fixture.

In my opinion this fixture is efficient in the way that you declare those variables once in the code run for Otel, this way they become the class variables and can be called by self instead of getting them each time in every function/fixture or test in the same code run.
If you notice I need to use these variables in different fixtures/functions many times.

If you have a better working suggestion to solve this issue let's talk about it cause I've tried not a few (by using the existing "instance_data") and this was the most efficient so far that could work well.

test_suite/package/otel_package/fixtures.py
```diff
@@ -0,0 +1,111 @@
...
+    self = request.node.cls
+    instance_data = helpers.instance_data(host)
+    self.instance_id = instance_data['instance_id']
+    self.instance_dns = instance_data['public_dns']
+    self.instance_region = instance_data['availability_zone'][:-1]
+
+
+@pytest.fixture(scope='function')
+def check_instance_status(request, host, timeout_seconds=300):
```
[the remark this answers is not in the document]
sshmulev: It's not only that, in additon I will have to declare the variable/s I want to use from this fixture here, for example:
self.instance_id = instance_data['instance_id']

Instead I'm doing this only **one time** in the "initialize_veribales" fixture that sets those variables to the class property which can be called by self anywhere in this code run. (including the test itself, and all the other fixtures within this run cause I'm using a request.node.cls that knows that this property belongs to Otel class.
In my opinion this solution is more appropriate for oop, but if you still think this is not a good idea let's talk about it and raise a better solution.

test_suite/package/otel_package/fixtures.py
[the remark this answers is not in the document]
narmaku: Yes, it makes sense. Thanks for the explanation. You're right, the code is more efficient this way.
The only thing is that `instance_data` fixture is used _everywhere_ on purpose, for metadata generation in the reports and it should be included on each test case, because this metadata is needed for analysis purposes.
One example is in the instance JSON metadata that can be seen here:
image

This metadata is also used by our analyzers to know about the context on which the test was run, and extract the data needed (e.g. the AMI ID, region, etc.).
The above are the real reasons why we use scope function and `autouse=True`. So, given that it's already there, I thought you can just use it instead of declaring the variables, which would be redundant.

Also I don't know the class variables approach would behave in the case we run tests in parallel on multiple instances. Although it is not the case for packages testing (yet?).

test_suite/generic/helpers.py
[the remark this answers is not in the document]
narmaku: Thanks, I think we can try with your proposal **n.1**, see if it works as expected, with multiple deployments.
If that does not behave correctly due to class instantiation in parallel testing scenarios, maybe we can just use `instance_data` fixture.

test_suite/package/otel_package/fixtures.py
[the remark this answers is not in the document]
sshmulev: If we change the scope of the fixture to 'class', it will be executed once per test class instead of once per test function.

test_suite/package/otel_package/fixtures.py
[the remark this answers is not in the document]
sshmulev: The dicsion was to migrate __get_instance_data_from_json to helpers.py since it is more suitable as a general function rather configuration one before the tests, and to use it directly in "initialize_variables" fixture.

test_suite/generic/helpers.py
```diff
@@ -1,10 +1,10 @@
...
 
-def instance_data(host):
-    values_to_find = [host.backend.hostname]
-    values_to_find.extend(host.addr(host.backend.hostname).ipv4_addresses)
-
-    return conftest.__get_instance_data_from_json(
-        key_to_find='address', values_to_find=values_to_find
-    )
+def __get_instance_data_from_json(key_to_find, values_to_find):
+    with open(os.environ['CIV_INSTANCES_JSON'], 'r') as f:
```
narmaku, line 6: We can make it even more generic and pass the json file path as function arg.

test_suite/generic/helpers.py
```diff
@@ -1,10 +1,10 @@
...
 
 
-def instance_data(host):
-    values_to_find = [host.backend.hostname]
-    values_to_find.extend(host.addr(host.backend.hostname).ipv4_addresses)
-
-    return conftest.__get_instance_data_from_json(
-        key_to_find='address', values_to_find=values_to_find
-    )
+def __get_instance_data_from_json(key_to_find, values_to_find):
```
narmaku, line 5:
```suggestion
def __get_instance_data_from_json(instances_json_path, key_to_find, values_to_find):
    with open(instances_json_path, 'r') as f:
```

narmaku (commented):
LGTM - just one comment to make the `__get_instance_data_from_json` function even more generic, now that it is in the helpers file.

test_suite/generic/helpers.py
sshmulev: I'm not sure I understand why this change is better than the existing one , I will have to pass more arguments to get the same instance data that we got before.
If we need some other function to make some specific operation I would suggest to do that separately and not mix it with this one that should only get the instance data, or maybe I just didn't understand the purpose of this idea?

narmaku: Ok, a few things:
- I didn't say it was better. It was just a suggestion.
- If it's a no-go for me I would have marked the pull requests as "needs work / request changes" again. But I just commented on it. So feel free to reject these comments without any further explanation, if that's the case.
- I think you understood the opposite. I was suggesting the change to make the function actually less specific, instead of having the env var reference inside there.
- This follows the "Separation of concerns" principle which is basically making sure the modules and functions only mind their business. A "helper" module shouldn't have this kind of references to env vars. Again, that's just a point of view, not a mandate.

In summary, if you think my comment does not make sense, you can always sync with me via chat, and feel free to also "resolve" the comments right away, when they don't indicate a showstopper.

narmaku:
Please resolve the old/outdated conversations, otherwise they flood the pull request history and it's harder to follow-up. Also I won't get notified about resolutions and my latest code review status won't change, which will be even more confusing. Thanks.

narmaku (approved):

test_suite/generic/helpers.py
sshmulev: Oh ok I didn't notice to this line "with open(instances_json_path, 'r') as f:"  this is why I didn't understand what's the benefit from this :) cause then it showed this one : "with open(os.environ['CIV_INSTANCES_JSON'], 'r') as f:" as it supposed to stay and not removed.

In general I agree about this, but since it is CIV Project so I don't think other instance json file should be handled here.

But to make it modular as you suggested I would create a configuration veriable in the helpers file since it's permenent variable for this CIV project and it's only 'CIV_INSTANCES_JSON', instead of adding this each time you call this function.

INSTANCES_JSON_PATH = <path/CIV_INSTANCES_JSON>
def __get_instance_data_from_json(path = 'INSTANCES_JSON_PATH', key_to_find, values_to_find):
        with open(instances_json_path, 'r') as f:
            instances_json_data = json.load(f)
            for instance in instances_json_data.values():
                if key_to_find in instance.keys() and instance[key_to_find] in values_to_find:
                    return instance

test_suite/generic/helpers.py
sshmulev: Oh ok I didn't notice to this line "with open(instances_json_path, 'r') as f:" this is why I didn't understand what's the benefit from this :) cause then it showed this one : "with open(os.environ['CIV_INSTANCES_JSON'], 'r') as f:" as it supposed to stay and not removed.

In general I agree about this, but since it is CIV Project so I don't think other instance json file should be handled here.

But to make it modular as you suggested I would create a configuration veriable in the helpers file since it's permenent variable for this CIV project and it's only 'CIV_INSTANCES_JSON', instead of adding this each time you call this function.

INSTANCES_JSON_PATH = <path/CIV_INSTANCES_JSON>
def __get_instance_data_from_json(path = 'INSTANCES_JSON_PATH', key_to_find, values_to_find):
with open(path, 'r') as f:
instances_json_data = json.load(f)
for instance in instances_json_data.values():
if key_to_find in instance.keys() and instance[key_to_find] in values_to_find:
return instance

narmaku: LGTM !

sshmulev: Fix is in this commit: https://github.com/osbuild/cloud-image-val/pull/316/commits/37a7f1e96de7ddaf921c3d710a975463d04626a3

Verified and it passed.
Ack if you approve this and merge:)

narmaku: Approved!

PR #325, opened 2024-06-20
Added awscli2 package test suite.

narmaku:
Related task: CLOUDX-927

### DIFF
--- a/test_suite/package/test_awscli2.py
+++ b/test_suite/package/test_awscli2.py
@@ -11,7 +11,7 @@ class TestsAwsCli2:
     @pytest.fixture(scope='module', autouse=True)
     def import_aws_credentials(self, host):
         # Generate temporary credentials for this test.
-        civ_local_command_to_run = 'aws sts get-session-token --duration-seconds 3600 --output json'
+        civ_local_command_to_run = 'aws sts get-session-token --duration-seconds 120 --output json'
 
         result = host.backend.run_local(civ_local_command_to_run)
         assert result.succeeded, \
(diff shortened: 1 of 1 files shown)


test_suite/package/test_awscli2.py
```diff
@@ -0,0 +1,76 @@
...
+from lib import test_lib
+
+
+@pytest.mark.package
+@pytest.mark.run_on(['>=rhel9.5'])
+class TestsAwsCli2:
+    @pytest.fixture(scope='module', autouse=True)
+    def import_aws_credentials(self, host):
+        # Generate temporary credentials for this test.
+        civ_local_command_to_run = 'aws sts get-session-token --duration-seconds 3600 --output json'
```
narmaku, line 14: Tests take less than 1 minute to run, so maybe we should reduce this number to something much smaller? E.g. `90`

ConorC117:
Code looks great to me! I had a good think about other worthwhile test cases to include, but to the best of my knowledge, I think the current tests are enough to ensure the package is working properly :)

test_suite/package/test_awscli2.py
ConorC117: I think reducing the number to something much smaller is a good idea too. I'd probably set it to 90 like you suggested or maybe 120 to be extra cautious!

narmaku:
```suggestion
        civ_local_command_to_run = 'aws sts get-session-token --duration-seconds 120 --output json'
```

PR #327, opened 2024-07-02
Removed Azure vhd uri name check to be more upstream-oriented.

narmaku:
Also made arch and storage_account mandatory fields for deployments that involve vhd uris.


narmaku:
PR in osbuild-composer CI so that it includes the needed information for deployments using vhd blobs:

- https://github.com/osbuild/osbuild-composer/pull/4239

**NOTE**: The PR above listed needs to be merged first.

PR #331, opened 2024-08-05
CIV: Refactor flaky TC of no failing serivces

sshmulev:
CLOUDX-592

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -471,8 +471,20 @@ class TestsServices:
                         f'rpm -qf "$(systemctl show --value --property=FragmentPath {service_name})"'
                     )
 
-            assert len(failing_services) == 0, \
-                f'There are failing services: {",".join(failing_services)}'
+            try:
+                assert len(failing_services) == 0, \
+                    f'There are failing services: {",".join(failing_services)}'
+            except AssertionError:
+                for service in failing_services:
+                    print(f"{service} is failing, attempting restart...")
+                    try:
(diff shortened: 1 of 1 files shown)


test_suite/generic/test_generic.py
```diff
@@ -471,8 +471,20 @@ def test_no_fail_service(self, host):
                         f'rpm -qf "$(systemctl show --value --property=FragmentPath {service_name})"'
                     )
 
-            assert len(failing_services) == 0, \
-                f'There are failing services: {",".join(failing_services)}'
+            try:
```
narmaku, line 474: Overall, just to be clear and avoid back and forth, this is the proposal I talked about, as a whole:

```
failed_services_after_restart = []
if len(failing_services) > 0:
    for service in failing_services:
        print(f"{service} is failing, attempting restart...")

        host.run(f"systemctl restart {service}")
        time.sleep(5)

        result = host.run(f"systemctl is-failed {service}")
        if 'fail' in result.stdout:
            print(f"Service {service} is still failing after restart")
            failed_services_after_restart.append(service)
        
assert len(failed_services_after_restart) == 0, \
    f'There are failing services: {",".join(failed_services_after_restart)}'
```

narmaku (changes requested):

test_suite/generic/test_generic.py
sshmulev: Yes sounds good!, I'll verify this one and push again, Thanks!

narmaku (approved):
LGTM!

PR #333, opened 2024-08-06
CIV: Refactor test_yum_group_install

sshmulev:
https://issues.redhat.com/browse/CLOUDX-837

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -558,12 +558,25 @@ class TestsAWS:
             pytest.skip('Not applicable to Atomic host AMIs')
 
         with host.sudo():
-            assert host.run_test('yum -y groupinstall "Development tools"'), \
-                'Error while installing Development tools group'
+            attempts = 2
+            for attempt in range(attempts):
+                try:
+                    assert host.run_test('yum -y groupinstall "Development tools"'), \
+                        'Error while installing Development tools group'
+                    break
+                except AssertionError as e:
+                    err_message = "This system is not registered to Red Hat Subscription Management"
+                    if err_message in str(e) and attempt == 0:
+                        host.run(
+                            'echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf'
+                            ' && yum clean all'
+                        )
+                    elif attempt == 1:
+                        raise AssertionError('Error while installing Development tools group after retrying')
 
-            package_to_check = 'glibc-devel'
-            assert host.package(package_to_check).is_installed, \
-                f'{package_to_check} is not installed'
+                package_to_check = 'glibc-devel'
+                assert host.package(package_to_check).is_installed, \
+                    f'{package_to_check} is not installed'
 
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel'])
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -558,25 +558,31 @@ class TestsAWS:
             pytest.skip('Not applicable to Atomic host AMIs')
 
         with host.sudo():
-            attempts = 2
-            for attempt in range(attempts):
-                try:
-                    assert host.run_test('yum -y groupinstall "Development tools"'), \
-                        'Error while installing Development tools group'
-                    break
-                except AssertionError as e:
-                    err_message = "This system is not registered to Red Hat Subscription Management"
-                    if err_message in str(e) and attempt == 0:
-                        host.run(
-                            'echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf'
-                            ' && yum clean all'
-                        )
-                    elif attempt == 1:
-                        raise AssertionError('Error while installing Development tools group after retrying')
+            dev_tools_install_command = 'yum -y groupinstall "Development tools"'
+            result = test_lib.print_host_command_output(host, dev_tools_install_command, capture_result=True)
 
-                package_to_check = 'glibc-devel'
-                assert host.package(package_to_check).is_installed, \
-                    f'{package_to_check} is not installed'
+            if result.failed:
+                err_message = "This system is not registered to Red Hat Subscription Management"
+                if err_message in result.stderr:
+                    print('"Development tools" installation attempt failed. Trying to apply a workaround...')
+                    host.run(
+                        'echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf'
+                        ' && yum clean all'
+                    )
+
+            result_second_attempt = test_lib.print_host_command_output(
+                host, dev_tools_install_command, capture_result=True
+            )
+
+            assert result_second_attempt.succeeded, (
+                'Error while installing Development tools group after two attempts. '
+                'Check test case output for more details.'
+            )
+            print('"Development tools" installed successfully.')
+
+            package_to_check = 'glibc-devel'
+            assert host.package(package_to_check).is_installed, \
+                f'{package_to_check} is not installed'
 
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel'])
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -572,8 +572,8 @@ class TestsAWS:
                     )
 
                     assert host.run(dev_tools_install_command).succeeded, (
-                        f'Error while installing Development tools'
-                        f' group after two attempts with error: {result.stderr}'
+                        f'Error while installing Development tools '
+                        f'group after two attempts with error: {result.stderr}'
                     )
                     print('"Development tools" installed successfully.')
 
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -561,7 +561,7 @@ class TestsAWS:
             dev_tools_install_command = 'yum -y groupinstall "Development tools"'
             result = host.run(dev_tools_install_command)
 
-            if result.exit_status != 0:
+            if result.failed:
                 print(f'Command faild with error on first attempt: {result.stderr}')
                 err_message = "This system is not registered to Red Hat Subscription Management"
                 if err_message in result.stderr:
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -575,7 +575,7 @@ class TestsAWS:
                         f'Error while installing Development tools '
                         f'group after two attempts with error: {result.stderr}'
                     )
-                    print('"Development tools" installed successfully.')
+            print('"Development tools" installed successfully.')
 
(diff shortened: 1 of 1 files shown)


test_suite/cloud/test_aws.py
```diff
@@ -558,12 +558,25 @@ def test_yum_group_install(self, host):
...
 
         with host.sudo():
-            assert host.run_test('yum -y groupinstall "Development tools"'), \
-                'Error while installing Development tools group'
-
-            package_to_check = 'glibc-devel'
-            assert host.package(package_to_check).is_installed, \
-                f'{package_to_check} is not installed'
+            attempts = 2
+            for attempt in range(attempts):
```
narmaku, line 562: As there will always be only two attempts, I would remove this for loop so we can reduce nesting.

test_suite/cloud/test_aws.py
```diff
@@ -558,12 +558,25 @@ def test_yum_group_install(self, host):
...
+                try:
+                    assert host.run_test('yum -y groupinstall "Development tools"'), \
```
narmaku, line 564: If we need to check for the result first, we can avoid try/except to capture the `AssertionError` we produced by ourselves.

Try/except mechanism should not be used to control the flow of expected/known errors, but to handle unexpected situations instead.

test_suite/cloud/test_aws.py
```diff
@@ -558,12 +558,25 @@ def test_yum_group_install(self, host):
...
+                        'Error while installing Development tools group'
+                    break
+                except AssertionError as e:
```
narmaku, line 567: This can be checked by using `host.run(command).stderr` instead.

test_suite/cloud/test_aws.py
narmaku, line 562: Proposal to reduce nesting and loops:

```
dev_tools_install_command = 'yum -y groupinstall "Development tools"'

result = test_lib.print_host_command_output(host, dev_tools_install_command, capture_result=True)

if result.failed:
    err_message = "This system is not registered to Red Hat Subscription Management"
    if err_message in result.stderr:
        print('"Development tools" installation attempt failed. Trying to apply a workaround...')

        host.run('echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf')
        host.run('yum clean all')

result_second_attempt = test_lib.print_host_command_output(host, dev_tools_install_command, capture_result=True)

assert result_second_attempt.succeeded, \
    'Error while installing Development tools group after two attempts. Check test case output for more details.'

print('"Development tools" installed successfully.')

package_to_check = 'glibc-devel'
assert host.package(package_to_check).is_installed, \
    f'{package_to_check} is not installed'

```

* `print_host_command_output`: To run the command in the host, and at the same time pretty-print the output (either if it's an error it will be pretty-printed in the test case output). By using `capture_result=True`, we make sure to retrieve the `CommandResult` object for further checks later in the code.
* Some additional print lines to inform us in the test case what happened and which steps took place during test runtime (for further debugging if needed).

narmaku (changes requested):

test_suite/cloud/test_aws.py
narmaku: How should we handle any other type of errors? I remember this test case was a "victim" of the DB_PAGE_NOT_FOUND issue when rpm database is broken after boot, for some reason.
In that case, should we do a second attempt? Or fail inmediately?

sshmulev: I agree, seems like a better approach!
Thanks

narmaku:
Please rebase from main and update the branch, so new runners will be used and CI tests will hopefully pass.

test_suite/cloud/test_aws.py
sshmulev: Regarding the prints, maybe we should use instead testing logs for the steps we do, I see that "print_divider" is something that is being used in CIV, what do you think?

sshmulev: Although the "print_divider" doesn't have the options to set logging levels like ( INFO/ ERRORs etc...)
I think it's a better practice for debugging

narmaku: Oh yes I forgot about that one! That's great for delimiting steps :+1:

test_suite/cloud/test_aws.py
```diff
@@ -558,8 +558,27 @@ def test_yum_group_install(self, host):
...
+            if result.failed:
+                err_message = "This system is not registered to Red Hat Subscription Management"
+                if err_message in result.stderr:
+                    print('"Development tools" installation attempt failed. Trying to apply a workaround...')
+                    host.run(
+                        'echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf'
+                        ' && yum clean all'
+                    )
+
+            result_second_attempt = test_lib.print_host_command_output(
```
narmaku, line 573: This should be indented inside the if statement. Maybe it was a mistake from my code suggestion.

test_suite/cloud/test_aws.py
```diff
@@ -558,8 +558,27 @@ def test_yum_group_install(self, host):
...
+                host, dev_tools_install_command, capture_result=True
+            )
+
+            assert result_second_attempt.succeeded, (
```
narmaku, line 577: This should be indented inside the if statement. Maybe it was a mistake from my code suggestion.

test_suite/cloud/test_aws.py
```diff
@@ -558,8 +558,27 @@ def test_yum_group_install(self, host):
...
+                'Error while installing Development tools group after two attempts. '
+                'Check test case output for more details.'
+            )
+            print('"Development tools" installed successfully.')
```
narmaku, line 581: This one is fine (outside the if statement).

narmaku (changes requested):

narmaku:
@sshmulev  build is failing, but because the test case is marked as "pub", so it will not be run in the Image Builder CI. We have a bug where CIV returns 1 when there were no test cases collected to run. **So we can ignore this.**

However, we should test in a few AMIs from a push task. Once we test that and we are sure the code behaves as expected, we can merge this.

test_suite/cloud/test_aws.py
sshmulev: I think I will create a separate PR to set the logging because the 'print devider' doesn't show the output in the console, which might also help when debugging just from Jenkins console output or locally.
There's a logging module that exist in Python, it's pretty common and it does the job: prints the output to console during run time and also in the final reports.

test_suite/cloud/test_aws.py
sshmulev: Oh that's right didn't notice this, thanks!

test_suite/cloud/test_aws.py
narmaku: We have a ticket in our backlog to implement better logging the whole project.
For now let's keep it simple and the current solution does the job as you said. :+1:

test_suite/cloud/test_aws.py
```diff
@@ -559,9 +559,10 @@ def test_yum_group_install(self, host):
 
         with host.sudo():
             dev_tools_install_command = 'yum -y groupinstall "Development tools"'
-            result = test_lib.print_host_command_output(host, dev_tools_install_command, capture_result=True)
+            result = host.run(dev_tools_install_command)
 
-            if result.failed:
+            if result.exit_status != 0:
```
narmaku, line 564: `result.failed` can still be used here.

test_suite/cloud/test_aws.py
```diff
@@ -570,13 +571,9 @@ def test_yum_group_install(self, host):
...
-                    )
-
-                    assert result_second_attempt.succeeded, (
-                        'Error while installing Development tools group after two attempts. '
-                        'Check test case output for more details.'
+                    assert host.run(dev_tools_install_command).succeeded, (
+                        f'Error while installing Development tools'
+                        f' group after two attempts with error: {result.stderr}'
                     )
                     print('"Development tools" installed successfully.')
```
narmaku, line 578: This print should be indented outside the two `if` statements.

narmaku (changes requested):
LGTM, just a small comment.

test_suite/cloud/test_aws.py
sshmulev: ack

test_suite/cloud/test_aws.py
sshmulev: ack

test_suite/cloud/test_aws.py
```diff
@@ -558,8 +558,24 @@ def test_yum_group_install(self, host):
...
+                    host.run(
+                        'echo -e "enabled=0" > /etc/yum/pluginconf.d/subscription-manager.conf'
+                        ' && yum clean all'
+                    )
+
+                    assert host.run(dev_tools_install_command).succeeded, (
+                        f'Error while installing Development tools '
+                        f'group after two attempts with error: {result.stderr}'
+                    )
+            print('"Development tools" installed successfully.')
```
narmaku, line 578: I think there's still an indentation issue here. We want to print this message even if installation succeeded after the first attempt (without having to enter into the if for a second attempt, applying workaround, etc.).

```suggestion
        print('"Development tools" installed successfully.')
```

So, by doing this, we will make sue that the `"Development tools" installed successfully.` message will be printed in all the "successful" scenarios (that is, with and without a retry).

narmaku (changes requested):

test_suite/cloud/test_aws.py
sshmulev: I don't see the issue you mentioned here, according to the existing code it does what you said.
It's outside the 'if' statement, so if that's not true (pass)it will be printed
and if it passes the assert inside this "if" it will be printed.

narmaku: Yeah, forget about what I said, it looks like the comment was added while you were pushing again, so it does not make sense any longer. Let's try to be more careful next time to avoid this situation where comments can be outdated with multiple pushes done in a short amount of time.

narmaku (approved):
LGTM

narmaku:
image
Output of successful scenario looks like expected! Merging.

test_suite/cloud/test_aws.py
```diff
@@ -558,8 +558,24 @@ def test_yum_group_install(self, host):
             pytest.skip('Not applicable to Atomic host AMIs')
 
         with host.sudo():
-            assert host.run_test('yum -y groupinstall "Development tools"'), \
-                'Error while installing Development tools group'
+            dev_tools_install_command = 'yum -y groupinstall "Development tools"'
+            result = host.run(dev_tools_install_command)
+
+            if result.failed:
```
lslebodn, line 564: IIUC it should fail just for BYOS due to missing subscription.
But it should work with PAYG due to rhui respositories.

@narmaku Shall we immediately fail for PAYG?

Or do I miss some context :-)

PR #335, opened 2024-08-11
Extend Opentelemetry-Collector testing with SELinux policies

sshmulev:
https://issues.redhat.com/browse/CLOUDX-1005

### DIFF
--- a/test_suite/generic/helpers.py
+++ b/test_suite/generic/helpers.py
@@ -1,5 +1,6 @@
 import os
 import json
+from lib import test_lib
 
 INSTANCES_JSON_PATH = os.environ['CIV_INSTANCES_JSON']
 
@@ -10,3 +11,14 @@ def __get_instance_data_from_json(key_to_find, values_to_find, path=INSTANCES_JS
     for instance in instances_json_data.values():
         if key_to_find in instance.keys() and instance[key_to_find] in values_to_find:
             return instance
+
+
+def check_avc_denials(host):
+    command_to_run = 'x=$(ausearch -m avc 2>&1 &); echo $x'
+    result = test_lib.print_host_command_output(host,
+                                                command_to_run,
+                                                capture_result=True)
+
+    no_avc_denials_found = 'no matches' in result.stdout
+
+    assert no_avc_denials_found, 'There should not be any avc denials (selinux)'
--- a/test_suite/package/otel_package/test_otel.py
+++ b/test_suite/package/otel_package/test_otel.py
@@ -30,8 +30,8 @@ class TestOtel:
 
     @pytest.mark.usefixtures(
         check_instance_status.__name__,
-        install_packages.__name__,
         setup_conf.__name__,
+        install_packages.__name__,
         modify_iam_role.__name__,
         start_service.__name__
     )
--- a/test_suite/package/otel_package/test_otel.py
+++ b/test_suite/package/otel_package/test_otel.py
@@ -10,7 +10,7 @@ from .fixtures import (
 
 
 @pytest.mark.package
-@pytest.mark.run_on(['>=rhel9.4'])
+@pytest.mark.run_on(['>=rhel9.5'])
 @pytest.mark.usefixtures(
     initialize_variables.__name__
 )
--- a/test_suite/package/otel_package/test_otel.py
+++ b/test_suite/package/otel_package/test_otel.py
@@ -2,7 +2,7 @@ import time
 import pytest
 from lib import console_lib
 from test_suite.generic import helpers
-from .fixtures import (
+from test_suite.package.otel_package.fixtures import (
     initialize_variables, check_instance_status,
     install_packages, setup_conf, modify_iam_role, start_service
 )
--- a/test_suite/package/otel_package/test_otel.py
+++ b/test_suite/package/otel_package/test_otel.py
@@ -1,5 +1,6 @@
 import time
 import pytest
+from packaging import version
 from lib import console_lib
 from test_suite.generic import helpers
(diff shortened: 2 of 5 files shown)


test_suite/package/otel_package/test_otel.py
```diff
@@ -1,14 +1,15 @@
 import time
 import pytest
 from lib import console_lib
+from test_suite.generic import helpers
```
narmaku, line 4:
```suggestion
from test_suite import helpers
```
Now that I think about this, maybe we should move the helpers outside of the "generic" dir?
The "generic" dir name actually refers to the fact that test cases inside there may apply to most/all cloud providers.

narmaku (changes requested):

narmaku:
@sshmulev have you tested this in a RHEL-9.5 AMI? Could you share the results?

test_suite/package/otel_package/test_otel.py
sshmulev: I don't see an issue with this being inside the generic dir, in fact the function we moved there from the first place "__get_instance_data_from_json" is being used by "instance_data" and not only for aws.
If we see more general functions in the "helpers" file that are meant to be only for one cloud provider, maybe we should consider separating this.

test_suite/package/otel_package/test_otel.py
```diff
@@ -1,14 +1,15 @@
...
 from .fixtures import (
     initialize_variables, check_instance_status,
     install_packages, setup_conf, modify_iam_role, start_service
 )
 
 
 @pytest.mark.package
-@pytest.mark.run_on(['rhel9.4'])
+@pytest.mark.run_on(['>=rhel9.4'])
```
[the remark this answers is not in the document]
sshmulev: I know that Otel was released from 9.4 this is why I added this from 9.4 and above.
Only the Selinux should be added in rhel9.5 and on.

test_suite/package/otel_package/test_otel.py
```diff
@@ -66,3 +68,6 @@ def test_otel(self, host):
 
         console_lib.print_divider("Check for error logs in aws cli logs")
         assert "Invalid user" in self.check_aws_cli_logs(host, self.instance_region)
+
+        if float(host.system_info.release) > 9.4:
```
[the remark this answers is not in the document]
sshmulev: ACK, thanks

test_suite/package/otel_package/test_otel.py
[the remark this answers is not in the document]
narmaku: Oh ok, thanks for clariying!

sshmulev:
> @sshmulev have you tested this in a RHEL-9.5 AMI? Could you share the results?

Not yet since it's still in development by the devs

test_suite/package/otel_package/test_otel.py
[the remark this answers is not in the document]
sshmulev: I verified this with Felix, apparently, Otel will be released in 9.5 so I changed it back together with the Selinux extention.

test_suite/package/otel_package/test_otel.py
```diff
@@ -69,5 +70,5 @@ def test_otel(self, host):
         console_lib.print_divider("Check for error logs in aws cli logs")
         assert "Invalid user" in self.check_aws_cli_logs(host, self.instance_region)
 
-        if float(host.system_info.release) > 9.4:
+        if version.parse(host.system_info.release) > version.parse('9.4'):
```
narmaku, line 73: I think this check is redundant, since the test case is already applicable only to RHEL-9.5 and later versions.

I suggest to remove the check and always run the check for avc denials.

narmaku (changes requested):

test_suite/package/otel_package/test_otel.py
sshmulev: ACK

test_suite/generic/helpers.py
```diff
@@ -10,3 +11,14 @@ def __get_instance_data_from_json(key_to_find, values_to_find, path=INSTANCES_JS
...
+
+def check_avc_denials(host):
+    command_to_run = 'x=$(ausearch -m avc 2>&1 &); echo $x'
+    result = test_lib.print_host_command_output(host,
+                                                command_to_run,
+                                                capture_result=True)
+
+    no_avc_denials_found = 'no matches' in result.stdout
+
+    assert no_avc_denials_found, 'There should not be any avc denials (selinux)'
```
lslebodn, line 24: ATM AVC are not collected in some cases in the `test_no_avc_denials`
IMHO it is allways useful to log AVC; even know ones which should be skipped.

One would be able to check whether previous AVCs are already fixed or whether new one are introduced.

so maybe it would be nice to have a feature to skip tests here (disabled by default)
e.g.
```
def check_avc_denials(host, skip_message=None):

//snip
if skip_message:
    pytest.skip(skip_message)
else:
    assert no_avc_denials_found, 'There should not be any avc denials (selinux)'
```

sshmulev: will create another PR for this change.

sshmulev:
**Vrified succesffuly:**
**rhel-9.5-nightly-x86_64:**
https://redhat.gitlab.io/-/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/7631072269/artifacts/report.html?sort=result

 **rhel-9.5-nightly-aarch64:** 
https://redhat.gitlab.io/-/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/7639223157/artifacts/report.html?sort=result

PR #338, opened 2024-08-27
Update rhel-ha-aws-check.sh in CIV

sshmulev:
https://issues.redhat.com/browse/CLOUDX-236

### DIFF
--- a/scripts/aws/rhel-ha-aws-check.sh
+++ b/scripts/aws/rhel-ha-aws-check.sh
@@ -25,9 +25,16 @@ RHELMAJOR=$(rpm -q --queryformat="%{VERSION}" \
 # fence-agents-aws - Fence agent for Amazon AWS
 #              pcs - Pacemaker Configuration System
 #  python-requests - HTTP library, written in Python (RHEL7 only)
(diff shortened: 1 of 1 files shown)


narmaku:
Comparing the full raw file against the one in main here, it is slightly different:
* https://github.com/liangxiao1/ami-val/blob/main/ami_val/scripts/rhel-ha-aws-check.sh

scripts/aws/rhel-ha-aws-check.sh
```diff
@@ -426,5 +447,11 @@ for HAPKG in ${HAPKGS}; do
     fi
 done
 
+# Install awscli from pip
```
narmaku, line 450: It's a removal, not an installation. The original script has a different text here. 

Can we just copy and paste the whole original script to make sure we replace all the needed parts?

https://github.com/liangxiao1/ami-val/blob/main/ami_val/scripts/rhel-ha-aws-check.sh

narmaku (changes requested):
Please use the original script content to fully replace ours.

narmaku (approved):
LGTM!

PR #340, opened 2024-09-03
Added logic for RHEL AWS Stratosphere images.

narmaku:
These changes are part of CLOUDX-993, to adapt our tooling to deal with recent changes that have merged all type of AMIs together in the same testing request.

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -4,13 +4,12 @@ import re
 import pytest
 from packaging import version
 
-from lib import test_lib
+from lib import test_lib, aws_lib
 
 
 @pytest.fixture
 def instance_data_aws_web(host):
-    instance_document_url = 'http://169.254.169.254/latest/dynamic/instance-identity/document'
-    return json.loads(host.check_output(f'curl -s {instance_document_url}'))
+    return aws_lib.get_aws_instance_identity_from_web(host)
 
 
 @pytest.fixture
@@ -553,6 +552,7 @@ class TestsAWS:
 
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel'])
+    @pytest.mark.usefixtures('rhel_aws_marketplace_only')
     def test_yum_group_install(self, host):
         if test_lib.is_rhel_atomic_host(host):
             pytest.skip('Not applicable to Atomic host AMIs')
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -35,7 +35,7 @@ class TestsGeneric:
         Check there is no avc denials (selinux).
         """
         if (host.system_info.distribution == 'rhel' and float(host.system_info.release) == 9.4) \
-           or (host.system_info.distribution == 'centos' and host.system_info.release == "9"):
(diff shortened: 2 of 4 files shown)


test_suite/generic/test_generic.py
```diff
@@ -647,10 +655,11 @@ def test_cloud_cfg_netdev_rhel9(self, host):
 
 @pytest.mark.pub
 @pytest.mark.order(3)
+@pytest.mark.usefixtures('exclude_on_rhel_aws_strato')
 class TestsYum:
     # TODO: confirm if this test needs to be deprecated
     @pytest.mark.run_on(['rhel', 'fedora'])
-    def test_yum_repoinfo(self, host):
+    def test_yum_repoinfo(self, host, instance_data):
```
narmaku, line 662: Remove unused `instance_data` reference: 
```suggestion
    def test_yum_repoinfo(self, host):
```

test_suite/cloud/test_aws.py
```diff
@@ -4,13 +4,19 @@
...
 
 @pytest.fixture
 def instance_data_aws_web(host):
-    instance_document_url = 'http://169.254.169.254/latest/dynamic/instance-identity/document'
-    return json.loads(host.check_output(f'curl -s {instance_document_url}'))
+    return aws_lib.get_aws_instance_identity_from_web(host)
+
+
+@pytest.fixture
+def rhel_aws_marketplace_only(host):
```
narmaku, line 16: Maybe this marker is redundant, since we already have one in the conftest.py file, that is actually more generic?

test_suite/cloud/test_aws.py
```diff
@@ -553,6 +559,7 @@ def test_dracut_conf_xen(self, host):
 
     @pytest.mark.pub
     @pytest.mark.run_on(['rhel'])
+    @pytest.mark.usefixtures('rhel_aws_marketplace_only')
```
lslebodn, line 562: I though you need opposite.
rhui works on 1p(golden images) because they have billing code.
But does not work for pre-release testing for 3p(stratosphere) because they are not yet associated with the AWS marketplace product.

narmaku: Thanks for the comment. The new `rhel_aws_marketplace_only` fixture introduced in this PR actually excludes Strato images.
We don't need "Strato-only" tests, but quite the opposite. MAybe the name is not the best, and it is also redundant compared to the fixture added in conftest.py, so we may probably remove this. Waiting for more feedback from cloudx peers :)

lslebodn: > MAybe the name is not the best

yeah; I was confused from the name and did not check the implementation :-)

test_suite/generic/test_generic.py
```diff
@@ -647,6 +648,7 @@ def test_cloud_cfg_netdev_rhel9(self, host):
 
 @pytest.mark.pub
 @pytest.mark.order(3)
+@pytest.mark.usefixtures('exclude_on_rhel_aws_strato')
```
lslebodn, line 651: Where it is defined?
I cannot see it anywhere

PR #341, opened 2024-09-04
Fix commit id according to osbuild-copmposer

sshmulev:
Commit id was chaned according to this:
https://github.com/osbuild/osbuild-composer/blob/main/Schutzfile


narmaku:
Closing this PR as it has been done here:
* https://github.com/osbuild/cloud-image-val/pull/343

PR #350, opened 2024-12-14
change instance type to t3.large

alexxa:
This is to accommodate capacity limits of new regions: ap-southeast-7 and mx-central-1


narmaku:
Hi  @alexxa ,
As discussed offline, this may have cost implications in other deployments that other teams use and in general.
This is an instance type we used to use and we actually changed it to `t3.medium` to be more cost-effective while still keeping good instance performance.
We can always control which instance type is going to be used directly in the `resources.json` file.

PR #352, opened 2025-01-16
Cloudx 1060

pugazhendhiramakrishnan08121985:
to be reviewed

### DIFF
--- a/test_suite/cloud/test_azure.py
+++ b/test_suite/cloud/test_azure.py
@@ -313,8 +313,17 @@ class TestsAzure:
             cert_found = False
 
             if test_lib.is_rhel_sap(host):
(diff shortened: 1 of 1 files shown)


alexxa:
@pugazhendhiramakrishnan08121985 Hey, thanks for the PR! 

Can you please update it, as \*base\* cert can exist only on non EUS/E4S releases, in the SAP case it's the final minor releases, e.g. RHEL 7.9, 8.10, 9.10, and so on.. If suddenly \*base\* cert is added to a any other minor (EUS/E4S) releases, than it would be a wrong configuration, and we can't release an image like that.

alexxa:
correction: similarly for other, non SAP RHELs, i.e.  \*base*\ cert can exist only on final minor releases.

PR #353, opened 2025-01-16
Added RSA SSH key type as default.

narmaku:
Added docstring to the function explaining the behavior. Formatted the file with auto-formatter.


ConorC117 (approved):
LGTM!

alexxa:
LGTM. The only kind suggestion is consider using [Imperative mood](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches#n307) in commit messages ;) It comes from 50/72 length recommendation, where 50 is for the message itself and 72 is for the commit body. Verb endings like -s, -ed, or -ing eat chars which can be used differently or dropped.

PR #354, opened 2025-01-21
Update package version in awscli2 test suite

knivnia:


test_suite/package/test_awscli2.py
```diff
@@ -32,7 +32,7 @@ def import_aws_credentials(self, host):
                 f'Could not configure temporary AWS credentials. Error: {result.stderr}'
 
     def test_awscli2_version(self, host):
-        expected_version = '2.15.31'
+        expected_version = '2.22.9'
```
alexxa, line 35: @knivnia can you share please, why is there a particular version? is there any dependency with OS version? can it be equal or greater than? when the version needs a change and how often?

narmaku: @knivnia is this new version applicable to both RHEL-10 and RHEL-9? In that case, looks good to me.

knivnia: @alexxa as discussed, the hardcoded version is to make sure that tests are run against particular version. awscli2 package updated quite often (more than once a week), but afaik we test it only if needed.

@narmaku no, the versions are different for RHEL-9 and RHEL-10. This was meant to be a quick fix to not block CTC2, maybe @F-X64 can give more info on this.

F-X64: @alexxa @narmaku We're currently in the process of releasing 2.22.9 on RHEL 9 and 10 (which means it's not in there yet but we expect it to be this version).

narmaku: Ok, give nthat it's only one test case from the whole test suite, we can still check the result of the rest test cases. Right? In that case, I would also recommend to make sure we check current system version and based on that, check specific version. It is also up to us (the team) to decide whether this test case still makes sense for us or not. Perhaps we can have an enhanced way of checking this as we were discussing with Conor and then during daily stand-ups. I would say that for now if this fix works, let's use it then.
@knivnia , you can test this changes by using our automation pipeline and setting the branch to be this PR:
image
https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/pipeline_schedules/1549645/edit?id=1549645

knivnia: @narmaku [tests](https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/pipelines/1637804762) passed

PR #357, opened 2025-02-05
add net.ifnames absence on rhel10 test

alexxa:
Closes [CLOUDX-979](https://issues.redhat.com/browse/CLOUDX-979)

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -254,6 +254,23 @@ class TestsGeneric:
 
             assert host.file(grub2_file).linked_to == linked_to
 
+    @pytest.mark.run_on(['rhel10'])
+    def test_net_ifnames_usage(self, host):
+        """
+        Check that net.ifnames=0 is not used as a kernel boot parameter
+        Jira: CLOUDX-979
+        """
+
+        kernel_boot_param = 'net.ifnames=0'
+        cmdline_file = '/proc/cmdline'
+        grub_default_file = '/etc/default/grub'
+
+        assert not host.file(cmdline_file).contains(kernel_boot_param), \
+            f'There is unexpected {kernel_boot_param} in kernel real-time boot parameters.'
+
+        assert not host.file(grub_default_file).contains(kernel_boot_param), \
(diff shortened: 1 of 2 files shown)


test_suite/generic/test_generic.py
```diff
@@ -254,6 +254,23 @@ def test_grub_config(self, host):
 
             assert host.file(grub2_file).linked_to == linked_to
 
+    @pytest.mark.run_on(['>=rhel10.0'])
+    def test_net_ifnames_usage(self, host):
+        """
+        Check that net.ifnames=0 is not used as a kernel boot parameter
```
narmaku, line 260: Add the associated JIRA ticket here.
JIRA: CLOUDX-XXX

narmaku (changes requested):

test_suite/generic/test_generic.py
```diff
@@ -254,6 +254,24 @@ def test_grub_config(self, host):
 
             assert host.file(grub2_file).linked_to == linked_to
 
+    @pytest.mark.run_on(['rhel10.0'])
+    @pytest.mark.jira_skip(["CLOUDX-994"])
```
narmaku, line 258: Maybe it was not clear enough, but we should skip the test that is related to CLOUDX-994 ticket. Not this test itself. This test will work well on rhel-10 since it's specifically done for rhel-10, so no issues here.

In summary: Remove this marker from this test and move it to `test_hostkey_permissions`.

test_suite/generic/test_generic.py
```diff
@@ -254,6 +254,24 @@ def test_grub_config(self, host):
 
             assert host.file(grub2_file).linked_to == linked_to
 
+    @pytest.mark.run_on(['rhel10.0'])
```
narmaku, line 257:
```suggestion
    @pytest.mark.run_on(['rhel10'])
```
If we specify only major version, it will apply to all RHEL-10.x minor versions too (which is our intention unless something changes in later minor releases, then we update the test case or marker accordingly).

narmaku (changes requested):

narmaku (approved):
LGTM!

Issue #360, opened 2025-02-11
CIV is failing in `osbuild-composer` CI on Azure with RHEL-9.6 nightly


thozza:
We are seeing
```
╷
│ Error: retrieving Shared Image Gallery (Subscription: "[MASKED]"
│ Resource Group Name: "travis-ci-osbuild-composer"
│ Gallery Name: "civ_gallery_60007"): unexpected status 404 (404 Not Found) with error: ResourceNotFound: The Resource 'Microsoft.Compute/galleries/civ_gallery_60007' under resource group 'travis-ci-osbuild-composer' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix
│ 
│ 
╵
Traceback (most recent call last):
  File "/opt/app-root/src/cloud-image-val.py", line 85, in <module>
    exit(cloud_image_validator.main())
  File "/opt/app-root/src/main/cloud_image_validator.py", line 57, in main
    self.cleanup()
  File "/opt/app-root/src/main/cloud_image_validator.py", line 120, in cleanup
    self.infra_controller.destroy_infra()
  File "/opt/app-root/src/cloud/opentofu/opentofu_controller.py", line 204, in destroy_infra
    raise Exception('tofu destroy command failed')
Exception: tofu destroy command failed
[2025-02-10T17:27:09+00:00] ❌ Failed V2
```
in https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9093895821#L2690 and it already happened multiple times. The test seem to pass, but the following cleanup fails.

thozza:
@narmaku could you please take a look? Thank you!

alexxa:
my bet would be that resources are destroyed in parallel and due to the broken dependencies some can't be removed.

narmaku:
@alexxa Yeah, it is possibly the case. Unfortunately, the AZ resources are so co-dependent...
We can first try:
- Updating OpenTotu to its latest version (in docker-compose.yaml)
- Updating AZ provider version

If the above does not resolve the issue, I would say let's introduce a retry mechanism to try to deal with it by ourselves (I would really love to avoid this).

PR about the points above: 
- https://github.com/osbuild/cloud-image-val/pull/363

narmaku:
@thozza ,  #363 has been merged. Can you retry your failing jobs and see if it fixed the issue?

thozza:
[...]

PR #361, opened 2025-02-11
[WIP] change tofu destroy to ordered

alexxa:
## Description

OpenTofu destroyes everything in parallel by default. It might cause failures on destroying resources. 
This PR destroys tofu resources in the dependency order (VM -> NIC -> IP -> SIV ->SI -> Subnet -> Vnet), adds retry mechanism, and adds a final clean up to ensure complete cleanup

Possible fix for #360 

## Changes
 - destroy_resources and destroy_infra functions
 - the order resources are destroyed to desired instead of parallel default
 - fix sufix type

### DIFF
--- a/cloud/opentofu/opentofu_controller.py
+++ b/cloud/opentofu/opentofu_controller.py
@@ -1,4 +1,5 @@
 import os
+import time
 import json
 import re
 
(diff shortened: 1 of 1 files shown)


alexxa:
* my bad, it overrides cloud agnostic function to be azure centric
* unit test change is missing.

PR #364, opened 2025-02-13
test keys permissions against OpenSSH version

alexxa:
## Description

Ensure permissions for SSH keys are checked against OpenSSH version

Close [CLOUDX-994](https://issues.redhat.com/browse/CLOUDX-994)

## Changes
 - test for hostkey permissions


test_suite/cloud/test_aws.py
```diff
@@ -367,20 +367,22 @@ def test_max_cstate_is_configured_in_cmdline(self, host):
                         f'{line} must not be specified in AMIs that are not SAP'
 
     @pytest.mark.run_on(['all'])
-    @pytest.mark.jira_skip(["CLOUDX-994"])
     def test_hostkey_permissions(self, host):
         """
         Check that ssh files permission set are correct.
         BugZilla 2013644
+        Ensure permissions are aligned with a distro and release version
+        ClOUDX-994
```
F-X64, line 375: Code changes look good to me.
Gives me some time to nit-pick ;) 
```suggestion
        CLOUDX-994
```

test_suite/cloud/test_aws.py
```diff
@@ -367,20 +367,22 @@ def test_max_cstate_is_configured_in_cmdline(self, host):
...
+        # Default permission for private keys
         expected_mode = 0o640
-        if host.system_info.distribution == 'fedora' and \
-                version.parse(host.system_info.release) >= version.parse('38'):
-            # On Fedora 38, ssh_keys group no longer exists and ssh-keygen no longer chmods to 640, see
-            # - https://src.fedoraproject.org/rpms/openssh/c/b615362fd0b4da657d624571441cb74983de6e3f?branch=rawhide
-            # - https://src.fedoraproject.org/rpms/openssh/c/7a21555354a2c5e724aa4c287b640c24bf108780?branch=rawhide
+
+        if host.system_info.distribution == 'fedora' or \
+                version.parse(host.system_info.release).major == 10:
```
F-X64, line 384: Probably never going to be an issue but the conditional "Fedora" or "every OS with Major version 10" is not the same as "every Fedora version >= 38" or "every RHEL version >= 10"

F-X64 (changes requested):

test_suite/cloud/test_aws.py
alexxa: no idea what was on F38, but on F41 ssh group is present. We test on latest Fedora, namely 41. Therefore no version for Fedora and we will see how it goes. 
Then, it's not only for RHEL, but for CentOS too. For now let's check specifically for version 10 because earlier versions have 644, and if it comes to >10, we will see how it's there.

F-X64: In that case just leave it as is. If the change is permanent in RHEL >=10 we should change this though

F-X64 (approved):
LGTM. Thanks for your explanation.

test_suite/cloud/test_aws.py
```diff
@@ -367,20 +367,22 @@ def test_max_cstate_is_configured_in_cmdline(self, host):
...
```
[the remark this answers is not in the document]
alexxa: @narmaku what about CentOS? the marker says it's for all distros.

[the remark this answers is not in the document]
narmaku: Good point, it should then be something like this: (trying to simplify the concatenation of conditions):

```
    distro = host.system_info.distribution
    major_version = version.parse(host.system_info.release).major
    if  distro == 'fedora':
        ...
    elif (distro == 'rhel' or distro == 'centos') and major_version == 10:
        ...
    else:
        ...
```

narmaku (approved):
LGTM!

PR #365, opened 2025-02-17
test /boot mount presence

alexxa:
## Description

Test the presence of the /boot mount on Fedora, RHEL and CentOS.

The /boot partition exists if
-   release version is 8.y and the architecture is aarch64
-   release version is 9.y
-   release version is10.y and it's Azure LVM
-   it's Fedora
In all other cases the /boot mount doesn't exist on a system.

Close [CLOUDX-930](https://issues.redhat.com/browse/CLOUDX-930) and [CLOUDX-980](https://issues.redhat.com/browse/CLOUDX-980)

## Changes
 - adds a new test case checking the /boot mount presence


alexxa:
@F-X64 @ConorC117 it seems it's about flake8 w503 or w504 conflict.. either the break line before or after a binary operator.. Either we exclude one of them in CI or please suggest an alternative solution maybe?

F-X64:
> @F-X64 @ConorC117 it seems it's about flake8 w503 or w504 conflict.. either the break line before or after a binary operator.. Either we exclude one of them in CI or please suggest an alternative solution maybe?

Yes that is correct. The [W504](https://www.flake8rules.com/rules/W504.html) and [W503](https://www.flake8rules.com/rules/W503.html)  rules contradict each other.
As the docs state, that's a known issue since 2016... my only recommendation is that you decide on one pattern, stick to it and add the other rule to your flake ignore file or use a tool such as pre-commit with an opinionated configuration in your github actions and locally instead of running the individual flake8 step.

ConorC117:
LGTM! Though I'd personally prefer to follow rule W504 instead of W503. I think it is more readable, but I am happy to proceed with either approach

ConorC117 (approved):

test_suite/generic/test_generic.py
```diff
@@ -254,6 +254,35 @@ def test_grub_config(self, host):
...
+        """
+
+        release_major = version.parse(host.system_info.release).major
+        is_aarch64 = host.system_info.arch == 'aarch64'
+        is_azure = instance_data['cloud'] == 'azure'
+        lvm_check = host.run("lsblk -f | grep LVM").rc == 0
+        is_fedora = host.system_info.distribution == 'fedora'
+
+        if (
+           (release_major == 8 and is_aarch64)
```
alexxa, line 277: @alexxa todo: add is_aws here, it was forgotten.

PR #366, opened 2025-02-18
ignore W504 in flake8

alexxa:
## Description

The PR #365 is blocked by flake8 complaining about either [W504](https://www.flake8rules.com/rules/W504.html) and [W503](https://www.flake8rules.com/rules/W503.html), a line break after or before binary operator. Those rules contradict each other. Therefore let's exclude  [W504](https://www.flake8rules.com/rules/W504.html) and [W503](https://www.flake8rules.com/rules/W503.html) and agree to use a line break after binary operators. 

## Changes
 - flake8 ignore rules in CI, ignoring W504

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,4 +30,4 @@ jobs:
         run: pytest --cov-report term-missing --cov=cloud/terraform --cov=result/ --cov=main/ --cov=test_suite/ test/
 
       - name: Python linter
-        run: flake8 --ignore=E501
+        run: flake8 --ignore=E501,W504


knivnia:
@alexxa why not ignore W503 to keep code more readable with all operators lined up (it is also mentioned in [PEP8](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator))? or does it require major refactoring?

alexxa:
updated the PR to ignore w503 as @ConorC117 and @knivnia suggested. Now waiting on CI, if it passes, I will merge it.

PR #367, opened 2025-02-24
align ci runners to osbuild-composer

alexxa:
## Description

This PR aligns runners with osbuild-composer ones as per [PR4626](https://github.com/osbuild/osbuild-composer/pull/4626)

Close [CLOUDX-1186](https://issues.redhat.com/browse/CLOUDX-1186)

## Changes
 - CI runners to match osbuild-composer ones as per [PR4616](https://github.com/osbuild/osbuild-composer/pull/4626)

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -29,14 +29,14 @@
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f40/f40-x86_64-updates-released-20250201"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f40/f40-x86_64-updates-released-20250225"
           }
         ],
         "aarch64": [
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f40/f40-aarch64-updates-released-20250201"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f40/f40-aarch64-updates-released-20250225"
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -66,6 +66,7 @@ prepare-rhel-internal-runners:
       - RUNNER:
           # NOTE: 1 runner prepares for all arches b/c subsequent jobs download
           # artifacts from all previous jobs and the last one wins
+          - aws/rhel-9.6-nightly-x86_64
           - aws/rhel-10.0-nightly-x86_64
```
alexxa, line 70: @narmaku Nico, osbuild-composer has [two runners](https://github.com/osbuild/osbuild-composer/blob/main/.gitlab-ci.yml#L181) here. Is there any reason why we have/had only one?

narmaku: Yes, back in the days with Juan (former Image Builder team member) we discovered that you only need one runner to prepare the internals for all versions and architectures.

So, we reached to the conclusion that we don't really need both to run, just one. However this information is from ~2 years ago and may be wrong. Better to double-check with someone from osbuild. CC @thozza

thozza: Ours `.gitlab-ci.yml` has different `rules` set for the job, see https://github.com/osbuild/osbuild-composer/blob/9b850de9f1f113744ad230420c2ce115f9afe610/.gitlab-ci.yml#L167-L168

As a result, only one runner is used, depending on the RHEL major version it runs for.
Examples:
9.6 nightly pipeline - https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1688676319
10.0 nightly pipeline - https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1688571275

So our case is different from yous, mainly because of the `rules`.

thozza:
BTW, this repo was supposed to get PRs with new rpmrepo snapshots automatically, but I found out that the bot that is making the PRs was missing a fork of this repository. This should be fixed now. I triggered a new snapshot to test it.

thozza:
> BTW, this repo was supposed to get PRs with new rpmrepo snapshots automatically, but I found out that the bot that is making the PRs was missing a fork of this repository. This should be fixed now. I triggered a new snapshot to test it.

https://github.com/osbuild/cloud-image-val/pull/368

alexxa:
Closing this one in favor of PR #368. @thozza many thanks for enabling automation for our repo 0/

PR #379, opened 2025-04-11
add bootc installed test

alexxa:
## Description

Test the installation of the system-reinstall-bootc package on 9.6 and 10.0

Close [CLOUDX-1267](https://issues.redhat.com/browse/CLOUDX-1267)

## Changes
 - adds a new test case checking the system-reinstall-bootc is installed

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -402,6 +402,21 @@ class TestsGeneric:
                 assert not host.file(file_to_check).exists, \
                     f'{file_to_check} should not exist in RHEL-8 and above'
 
+    @pytest.mark.run_on(['>=rhel9.6', 'rhel10'])
+    def test_bootc_installed(self, host):
+        """
+        Verify the system-reinstall-bootc package is installed
+        JIRA: CLOUDX-1267
+        """
+
+        with host.sudo():
+            print('rpm -q output: ')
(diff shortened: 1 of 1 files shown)


F-X64 (approved):
Looks good to me!

test_suite/generic/test_generic.py
```diff
@@ -400,6 +400,17 @@ def test_langpacks_conf(self, host):
...
+    @pytest.mark.run_on(['>=rhel9.6', 'rhel10'])
+    def test_bootc_installed(self, host):
+        """
+        Verify the system-reinstall-bootc package is installed
+        JIRA: CLOUDX-1267
+        """
+
+        with host.sudo():
+            assert host.package("system-reinstall-bootc").is_installed, \
+                'Expected system-reinstall-bootc package not installed'
```
F-X64, line 412: Log messages could potentially be more explicit. E.g. "System-reinstall-bootc package expected to be installed in RHEL >= 9.6, 10.0". I know that the test case is only executed on those versions, but I don't know if this is visible from the log output alone.  WDYT?

alexxa:
The CI probably fails because it's not in image-builder-crc repo, however it's in the latest builds in 9.6 and 10.0
We need to QA those advisories with this test being there, therefore merging.

PR #384, opened 2025-04-18
add keyfile plugin check for rhel10 and centos10

alexxa:
This changes a `test_eth0_network_adapter_setup` test to check for the NetworkManager keyfile plugin instead of the cfcfg file on RHEL10 and CentOS10.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -748,23 +748,32 @@ class TestsNetworking:
         """
         Make sure that eht0 default adapter is correctly setup:
             1. NETWORKING=yes in /etc/sysconfig/network
-            2. DEVICE=eth0 in /etc/sysconfig/network-scripts/ifcfg-eth0
+            2. For major_release<10: eth0 in /etc/sysconfig/network-scripts/ifcfg-eth0
+            3. For major_release>=10: exists /etc/NetworkManager/system-connections/*.nmconnection
 
         Does not apply to >fedora35: https://fedoramagazine.org/converting-networkmanager-from-ifcfg-to-keyfiles/
         """
         if instance_data['cloud'] == 'azure' and \
                 version.parse(host.system_info.release).major == 9:
             pytest.skip('Skipping due to cloud-init known issue in RHEL-9.x. See COMPOSER-2437 for details.')
+
         device_name = 'eth0'
+        device_config_path = f'/etc/sysconfig/network-scripts/ifcfg-{device_name}'
+        keyfile_plugin = '/etc/NetworkManager/system-connections/*.nmconnection'
 
         with host.sudo():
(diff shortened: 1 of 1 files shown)


test_suite/generic/test_generic.py
```diff
@@ -746,23 +746,32 @@ def test_eth0_network_adapter_setup(self, host, instance_data):
...
 
-            assert host.file(device_config_path).contains(f'^DEVICE=[{device_name}|\"{device_name}\"]'), \
-                f'Unexpected device name. Expected: "{device_name}"'
+            if release_major < 10:
+                assert host.file(device_config_path).contains(f'^DEVICE=[{device_name}|\"{device_name}\"]'), \
+                    f'Unexpected device name. Expected: "{device_name}"'
+            else:
+                keyfile = host.check_output(f"ls {keyfile_plugin} 2>/dev/null || true")
+                assert keyfile != "", \
+                    f'There is no keyfile plugin as "{keyfile_plugin}"'
```
F-X64, line 774: Typo (?)
Do you mean "in"?
```suggestion
                    f'There is no keyfile plugin in "{keyfile_plugin}"'
```

test_suite/generic/test_generic.py
```diff
@@ -746,23 +746,32 @@ def test_eth0_network_adapter_setup(self, host, instance_data):
...
                 'Invalid networking setup'
 
-            device_config_path = f'/etc/sysconfig/network-scripts/ifcfg-{device_name}'
+            release_major = version.parse(host.system_info.release).major
 
-            assert host.file(device_config_path).contains(f'^DEVICE=[{device_name}|\"{device_name}\"]'), \
-                f'Unexpected device name. Expected: "{device_name}"'
+            if release_major < 10:
+                assert host.file(device_config_path).contains(f'^DEVICE=[{device_name}|\"{device_name}\"]'), \
+                    f'Unexpected device name. Expected: "{device_name}"'
```
F-X64, line 770: In RHEL < 10 we've been checking the device configuration for the presence of the expected device.
Is there any content verification necessary for the keyfile in RHEL >= 10?

test_suite/generic/test_generic.py
alexxa: @F-X64 here we check a presence of any file matching a pattern, therefore "as".

test_suite/generic/test_generic.py
alexxa: what about adding it in Jira to investigate (as noted earlier I will open all Jira for these PRs retrospectively), and for now approving it as it is, so we have it updated for RHEL10 / not to block a CI just because of ifcfg -> keyflile plugin change?

F-X64 (approved):

PR #389, opened 2025-04-22
remove net.ifnames=0 check for rhel10

alexxa:
This is to fix a test which fails for RHEL10 on AWS as net.ifnames=0 is deprecated there.
In the future it's desired to unite two test on net.ifnames=0 as reported in [CLOUDX-1303](https://issues.redhat.com/browse/CLOUDX-1303)


F-X64 (approved):
LGTM

PR #390, opened 2025-04-22
change root lock test for rhel10

alexxa:
In RHEL10, passwd lock is noted as "L", not "LK" as in earlier releases.


F-X64 (approved):
LGTM

PR #391, opened 2025-04-22
Fix type mismatch in get_skip_vars

F-X64:
skip_aws, skip_azure, skip_gcp are expected to be strings not booleans.
-----------

The existing implementation is troublesome as it uses a combination of shell scripts and python code.
There is no type safety in place that ensures types are not mismatched which is reason enough for a refactoring of the existing code.

Example implementation for this code:
```
def get_skip_vars() -> Dict[str, str]:
    files_changed : List[str] = get_files_changed()

    if any("test_suite/generic/" in file for file in files_changed):
        return {cloud: "false" for cloud in ["skip_aws", "skip_azure", "skip_gcp"]}

    return {f"skip_{cloud}": "true" if f"test_suite/cloud/test_{cloud}.py" in files_changed else "false"
            for cloud in ["aws", "azure", "gcp"]}
```

### DIFF
--- a/schutzbot/get_civ_config.py
+++ b/schutzbot/get_civ_config.py
@@ -12,6 +12,7 @@ import sys
 import yaml
 
 from pprint import pprint
+from typing import Dict, List
 
 
 DEFAULT_CIV_CONFIG_PATH = "/tmp/civ_config.yaml"
@@ -165,23 +166,23 @@ def get_modified_methods_str():
     return " or ".join(list(modified_methods))
 
 
-def get_skip_vars():
-    skip_vars = {
+def get_skip_vars() -> Dict[str, str]:
+    skip_vars: Dict[str, str] = {
         "skip_aws": "false",
         "skip_azure": "false",
         "skip_gcp": "false"
     }
 
-    files_changed = get_files_changed()
+    files_changed: List[str] = get_files_changed()
 
     for file in files_changed:
         if "test_suite/generic/.*" in file:
             return skip_vars
 
     return {
-        "skip_aws": "test_suite/cloud/test_aws.py" in files_changed,
-        "skip_azure": "test_suite/cloud/test_azure.py" in files_changed,
-        "skip_gcp": "test_suite/cloud/test_gcp.py" in files_changed,
+        "skip_aws": "false" if "test_suite/cloud/test_aws.py" in files_changed else "true",
+        "skip_azure": "false" if "test_suite/cloud/test_azure.py" in files_changed else "true",
+        "skip_gcp": "false" if "test_suite/cloud/test_gcp.py" in files_changed else "true",
     }
 
 


schutzbot/get_civ_config.py
```diff
@@ -165,23 +166,23 @@ def get_modified_methods_str():
...
 
     for file in files_changed:
         if "test_suite/generic/.*" in file:
             return skip_vars
 
     return {
-        "skip_aws": "test_suite/cloud/test_aws.py" in files_changed,
-        "skip_azure": "test_suite/cloud/test_azure.py" in files_changed,
-        "skip_gcp": "test_suite/cloud/test_gcp.py" in files_changed,
+        "skip_aws": "false" if "test_suite/cloud/test_aws.py" in files_changed else "true",
```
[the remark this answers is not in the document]
alexxa: to me it seems, we should skip_<cloud> only if a corresponding file was NOT changed. If it was changed, we do need to ensure that updated tests executed successfully on given runners. Therefore, it seems the original logic was correct.

alexxa:
LGTM, just skip logic is questionable.

schutzbot/get_civ_config.py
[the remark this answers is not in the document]
narmaku: Yeah it makes sense to either revert this, or add "not" to each check inside the dict. Sorry about the mess. (@alexxa maybe this is why the tests were running much faster).

schutzbot/get_civ_config.py
```diff
@@ -165,23 +166,23 @@ def get_modified_methods_str():
...
         "skip_azure": "false",
         "skip_gcp": "false"
     }
 
-    files_changed = get_files_changed()
+    files_changed: List[str] = get_files_changed()
 
     for file in files_changed:
         if "test_suite/generic/.*" in file:
             return skip_vars
```
alexxa, line 181: @F-X64 @narmaku this PR doesn't change any test_suite/cloud/*.py files, meanwhile, the test_suite are executed, e.g. see [aws-rhel-9.4-ga-x86](https://gitlab.com/redhat/services/products/image-builder/ci/cloud-image-val-ci/-/jobs/9789785811#L2384), meaning skip variable in lines 182-184 is not returned, but it happens much earlier, on likes 180, and it's take as "false" from line 170, is it? Otherwise, why were test triggered? 

Then, on line 179, why do we check if test_suite/generic is changed and then return skip_vars? this logic I don't follow completely.

narmaku: Let's cover this in our pair programming call today.

schutzbot/get_civ_config.py
narmaku, line 183: Oh dear, now I see what happened. I was storing booleans and with the inverted logic here, instead of strings.

We need to rework this and simplify things...

narmaku: Would a revert be better in this case?
At this point I'm not sure if I broke things up due to my refactoring :/.
In the reverted logic, we can just nest some things inside an `if` statement.
The code will still be hard to maintain though (maybe harder than in this refactored version).

F-X64:
Closed PR as we've decided to revert the changes that caused this issue instead #399

PR #393, opened 2025-04-23
dhcp-client removal never needed for azure

alexxa:
By mistake I copy pasted removal of the dhcp-client on Azure, while it was for AWS only, and in general dhcp-client never was on Azure, therefore taking it back.


F-X64 (approved):
Good catch.

PR #394, opened 2025-04-23
ignore avc denial for irqbalance

alexxa:
There is a known issue with avc denial in irqbalance (RHEL-78630), which we need to ignore for now for our and osbuild-composer CIs to pass.


alexxa:
yes, tracked in CLOUDX-1304

PR #397, opened 2025-04-28
add gen1 and gen2 launch logic for Azure VMs

alexxa:
This contributes to [CLOUDX-1292](https://issues.redhat.com/browse/CLOUDX-1292)
However, this introduces only logic to launch VMs of both gen1 and gen2 types, if 'instance_type' is missing. Corresponding suffixes are added to the resources naming for easier debugging, as illustrated

Screenshot From 2025-04-28 20-59-16

In a separate PR there should be addressed test cases specific to gen1/gen2.

### DIFF
--- a/cloud/opentofu/azure_config_builder_v2.py
+++ b/cloud/opentofu/azure_config_builder_v2.py
@@ -13,7 +13,10 @@ class AzureConfigBuilderV2(BaseConfigBuilder):
         "azurerm": {"source": "hashicorp/azurerm", "version": f"~> {provider_version}"}
     }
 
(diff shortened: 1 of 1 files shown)


alexxa:
@F-X64 it's fine for review, but without our CI I don't feel comfortable on merging it. ATM, it's possible to asses production releases locally only, without stage ones.

F-X64:
> @F-X64 it's fine for review, but without our CI I don't feel comfortable on merging it. ATM, it's possible to asses production releases locally only, without stage ones.

Understood. I'll see to get the CI going as soon as possible.

Issue #398, opened 2025-04-29
[RHEL-8.10] TestsAuthConfig.test_smartcard_auth fails with CDN content


thozza:
The `TestsAuthConfig.test_smartcard_auth` test is failing in osbuild-composer RHEL-8.10 GA pipelines: 

- https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9857133616
- https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9857163531

In my opinion, this is caused by a change in the RPM that ships the config in RHEL-8, which falls outside the responsibility of the Image Builder team. The test case likely needs to be adjusted to reflect the current state of the configuration shipped in the RHEL-8.10 RPMs.

```
=================================== FAILURES ===================================
_ TestsAuthConfig.test_smartcard_auth[paramiko://civ-vm-image-rhel-810-x86-64-main-9857133616-11036.westeurope.cloudapp.azure.com] _
self = <test_generic.TestsAuthConfig object at 0x7f6ebc160670>
host = <testinfra.host.Host paramiko://civ-vm-image-rhel-810-x86-64-main-9857133616-11036.westeurope.cloudapp.azure.com>
    @pytest.mark.exclude_on(['>=rhel10.0'])
    def test_smartcard_auth(self, host):
        """
        Check file /etc/pam.d/smartcard-auth
        Bugzilla: 1983683
        """
    
>       self.__check_pam_d_file_content(host, 'smartcard-auth')
test_suite/generic/test_generic.py:941: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <test_generic.TestsAuthConfig object at 0x7f6ebc160670>
host = <testinfra.host.Host paramiko://civ-vm-image-rhel-810-x86-64-main-9857133616-11036.westeurope.cloudapp.azure.com>
file_name = 'smartcard-auth'
    def __check_pam_d_file_content(self, host, file_name):
        system_release_major_version = version.parse(host.system_info.release).major
        local_file = f'data/generic/{file_name}_rhel{system_release_major_version}'
        file_to_check = f'/etc/pam.d/{file_name}'
    
>       assert test_lib.compare_local_and_remote_file(host, local_file, file_to_check), \
            f'{file_to_check} has unexpected content'
[...]

PR #399, opened 2025-04-29
Revert "Add Gitlab CI variables."

F-X64:
Reverts osbuild/cloud-image-val#378 as we are currently unclear on its impact on functionality outside of the get_civ_config changes.


alexxa:
@F-X64 please update the PR title if it's not a Draft anymore. 
.gitlab-ci.yml.swp was committed probably by mistake. 
Anything else seems to be fine, let's wait on CI to complete.

F-X64:
Ah yes, the swp file was dropped during the original PR but there's no reason to add it again.

alexxa:
There is just one failure and it was already reported by @thozza #398, therefore merging it. 
If we would have CI working earlier, #398 wouldn't happen. 
Thanks, Felix 0/

alexxa (approved):

PR #410, opened 2025-05-19
Major refactoring to make CI/CD more modular.

narmaku:
- Add CI variables to easily run pipelines from Web UI.
- Modularize gitlab-ci yaml files for easier maintenance.


schutzbot/get_civ_config.py
narmaku, line 1: Now this file is left almost untouched and just added the essential if statement to avoid affecting Image Builder pipelines. IT should be 100% transparent for them.

narmaku:
Same as https://github.com/osbuild/cloud-image-val/pull/378 :) without the refactoring of get_civ_config file, which was causing issues last time. The rest is exactly the same, and should not cause any issues.

F-X64 (changes requested):
This split was a good idea, thanks for implementing it.
Just a couple of typos, other than that it looks good.
Also, I'm in favor of not introducing too many changes in get_civ_config.py in the same PR.

PR #412, opened 2025-05-22
test /boot size being >=960MiB

alexxa:
If /boot exists, it should be at least 960MiB as a lower threshold of 1024Mib


F-X64 (approved):

alexxa:
Merge when [composer PR1557](https://github.com/osbuild/images/pull/1557) is merged

PR #416, opened 2025-06-02
Add attach feature.

narmaku:
This feature allows you to reconnec to instances deployed by a previous run in the event CIV exited abruptly or environment cleanup was cancelled via CLI.

### DIFF
--- a/cloud-image-val.py
+++ b/cloud-image-val.py
@@ -8,66 +8,109 @@ from lib import console_lib
 
 parser = ArgumentParser(formatter_class=RawTextHelpFormatter)
 
-parser.add_argument('-r', '--resources-file',
-                    help='Path to the resources JSON file that contains the Cloud provider and the images to use.\n'
-                         'See cloud/sample/resources_<cloud>.json to know about the expected file structure.',
-                    default=None)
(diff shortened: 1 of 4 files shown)


F-X64 (changes requested):
I really like the idea, do you know in what timeframe are those instances kept alive?
I'm curious if it would make sense to check for the exit state of the CIV run and extend this into an auto retry logic in the future.
----
The implementation looks good, thanks.
Only added some requests for typos.

narmaku:
> I really like the idea, do you know in what timeframe are those instances kept alive? I'm curious if it would make sense to check for the exit state of the CIV run and extend this into an auto retry logic in the future.
> 
> The implementation looks good, thanks. Only added some requests for typos.

Hi @F-X64 ! Thanks for the review.
The instances are kept alive as long as [cloud-cleaner](https://github.com/osbuild/cloud-cleaner/actions/workflows/run_cloudx.yml) allows (I think it's 6 hours by default). In the other hand, once you re-attach, tests will re-run and then instances will be removed as usual. If you use `-s` (stop cleanup) option, it will ask you before removing the instances.

F-X64 (approved):

PR #534, opened 2026-03-17
test/generic: update list of authselect config files for RHEL 10 [HMS-9868]

achilleas-k:
I thought there would be more changes needed here.  I went through all the CIV-related failures from [this pipeline run](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/2359353652) (from osbuild/osbuild-composer#5021) and they're all because of the new authselect file that appeared in RHEL 10.


F-X64 (dismissed):
Aside from the osbuild install failures (I assume this is related to the version mismatch in the schutzfile) the pipeline looks good.

achilleas-k:
ah, let me bump that real quick.

F-X64 (approved):
Great work, thanks for getting this fixed!

PR #552, opened 2026-05-31
Add oci provider

sshmulev:
**Add OCI as a supported cloud provider**

This PR adds OCI support to CIV following the same pattern as AWS and Azure:
1. OCIConfigBuilder dynamically generates OpenTofu .tf.json config (VCN, subnet, internet gateway, route table, instance)
2. get_instances_oci() parses tofu show output into CIV's standard instance dict
3. OCI is registered in OpenTofuConfigurator and BaseConfigBuilder
4. cloud/sample/resources_oci.json provides a sample resources file
5. test_suite/cloud/test_oci.py adds OCI-specific tests (IMDS, cloud-init, networking, chrony)

Tested against a RHEL 9.8 custom image in OCI San Jose region.


cloud/opentofu/oci_config_builder.py
```diff
@@ -0,0 +1,167 @@
+from cloud.opentofu.base_config_builder import BaseConfigBuilder
+
+
+class OCIConfigBuilder(BaseConfigBuilder):
+    cloud_name = 'oci'
+    cloud_provider_definition = {'oci': {'source': 'oracle/oci', 'version': '~> 6.0'}}
```
F-X64: Is there a specific reason why we're using version ~> 6.0? Latest release would be at 8.16.
If there aren't any breaking changes we're concerned with I would advise against using something that hasn't been updated since April 2025.

cloud/opentofu/oci_config_builder.py
```diff
@@ -0,0 +1,167 @@
...
+
+        return self.providers_tf
+
+    def __get_all_regions_from_resources_file(self):
+        regions = [i['region'] for i in self.resources_dict['instances']]
+        return list(dict.fromkeys(regions))
+
+    def __new_oci_provider(self, region):
+        return {
+            'config_file_profile': 'DEFAULT',
```
F-X64: This might be an issue if you have multiple OCI profiles configured in your CLI.
Either make this configurable or document that users of CIV must use their DEFAULT profile for cloud image verification tasks.

cloud/opentofu/oci_config_builder.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        declared_subnet_id = f'oci_core_subnet.{instance["subnet"]}.id'
+
+        new_instance = {
+            'provider': f'oci.{region}',
+            'availability_domain': f'${{{declared_ad}}}',
+            'compartment_id': instance['compartment_id'],
+            'display_name': name_value,
+            'shape': instance.get('shape', 'VM.Standard.E4.Flex'),
+            'shape_config': {
```
F-X64: Always including the shape_config is fine for the Standard.E4 but not all instance types on OCI allow for it. This might be an issue if we run verifications on releases that are not certified for a .Flex type (e.g. some HPC or SAP releases might have that issue).
In that case we need to overwrite the shape and shape_config. 

You could of course also use a conditional here e.g.

```python
   if '.Flex' in new_instance['shape']:
      new_instance['shape_config'] = {
          'ocpus': instance.get('ocpus', 2),
          'memory_in_gbs': instance.get('memory_in_gbs', 16),
      }
```

F-X64 (changes requested):
@sshmulev Thank for the implementation! The overall structure is good and aligns with what we have for AWS and Azure. There are a couple of issues that need fixing / discussion before we move on though (see comments).

Please also update the PR with a proper description of your changes and their reasoning.

cloud/opentofu/oci_config_builder.py
sshmulev: That's right. I'll change to 8.0 to be on the safe side? what do you say?

F-X64: Depends. What was the reason you chose 6.0?

sshmulev: I was just mirroring the existing AWS/Azure architecture as a baseline, so no strict reason for 6.0. Since it's a fresh integration

F-X64: In that case please look for any possible breaking changes (so far I don't see anything) and update the version to the latest stable release.

F-X64 (commented):
LGTM

PR #568, opened 2026-07-01
Fix AWS test failures - Update tests for current image state

sshmulev:
Fixed four test failures that were caused by automation drift:

1. test_authselect_conf_files (AWS):
   - Removed switchable-auth from AWS expected files
   - switchable-auth introduced in authselect 1.7.0+ (Jan 2025)
   - Not present in current RHEL 9.8 AWS images
   - Tested on ami-0bb64587efd91f63e (RHEL 9.8 us-east-2): PASSED

2. test_iommu_strict_mode (AWS ARM):
   - Added version-specific logic for RHEL 9/10 ARM
   - RHEL 9/10: CONFIG_IOMMU_DEFAULT_DMA_STRICT not set, parameter not required
   - RHEL 8: CONFIG_IOMMU_DEFAULT_DMA_STRICT='y', parameter still required
   - Based on CLOUDX-1517 and RHEL-115405

3. test_subscription_manager_auto:
   - Added check for "Overall Status: Registered" status text
   - RHEL 10+ auto-registration v2 changed subscription-manager status output
   - No longer shows "Red Hat Enterprise Linux" or "Simple Content Access"
   - Now shows "Overall Status: Registered" instead
   - Tested on ami-00487d5140419958a (RHEL 10 eu-west-1): PASSED

4. test_yum_group_install cleanup:
   - Added pytest finalizer to remove "Development tools" packages after test
   - Prevents test pollution affecting test_unwanted_packages_are_not_present
   - alsa-lib is NOT in base images but gets installed as dependency
   - Verified: fresh instance has no alsa-lib, but appears after groupinstall
   - Tested: both tests now pass in sequence with cleanup

### DIFF
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -87,24 +87,30 @@ class TestsAWS:
     @pytest.mark.run_on(['rhel'])
     def test_iommu_strict_mode(self, host):
         """
-        Use "iommu.strict=0" in ARM AMIs to get better performance.
-        BugZilla 1836058
-        CLOUDX-1517, RHEL-115405
+        Checks the status of the "iommu.strict=0" kernel option on RHEL AMIs.
+        The option should be:
+        - Not present on x86_64 images (all versions)
+        - Present on RHEL 8 ARM images (BugZilla 1836058)
+        - Not present on RHEL 9+ ARM images (CLOUDX-1517, RHEL-115405)
         """
         option = 'iommu.strict=0'
         release_major = version.parse(host.system_info.release).major
+        arch = host.system_info.arch
 
         with host.sudo():
             iommu_option_present = host.file('/proc/cmdline').contains(option)
 
-            if host.system_info.arch == 'x86_64':
+            if arch == 'x86_64':
                 assert not iommu_option_present, f'{option} must not be present in x86_64 AMIs'
-            elif release_major >= 9:
-                assert not iommu_option_present, \
-                    f'{option} must not be present in RHEL {release_major} ARM AMIs (CLOUDX-1517)'
+            elif arch == 'aarch64':
+                if release_major >= 9:
+                    assert not iommu_option_present, \
+                        f'{option} must not be present in RHEL {release_major} ARM AMIs (CLOUDX-1517)'
+                else:
+                    assert iommu_option_present, \
+                        f'{option} must be present in RHEL {release_major} ARM AMIs (BugZilla 1836058)'
             else:
-                assert iommu_option_present, \
-                    f'{option} must be present in RHEL {release_major} ARM AMIs'
+                pytest.fail(f'Unexpected architecture: {arch}. This test only supports x86_64 and aarch64.')
 
     @pytest.mark.run_on(['rhel'])
     def test_unwanted_packages_are_not_present(self, host):
--- a/test_suite/cloud/test_aws.py
+++ b/test_suite/cloud/test_aws.py
@@ -89,16 +89,22 @@ class TestsAWS:
         """
         Use "iommu.strict=0" in ARM AMIs to get better performance.
         BugZilla 1836058
+        CLOUDX-1517, RHEL-115405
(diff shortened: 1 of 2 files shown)


Thilak-KN:
> 2. test_iommu_strict_mode (AWS ARM):

I verified this behaviour in RHEL-8,9,10 in AWS, and the changes in the test-case looks good to me.

@achilleas-k probably can review and approve the other changes.

test_suite/cloud/test_aws.py
```diff
@@ -89,16 +89,22 @@ def test_iommu_strict_mode(self, host):
         """
         Use "iommu.strict=0" in ARM AMIs to get better performance.
         BugZilla 1836058
+        CLOUDX-1517, RHEL-115405
```
achilleas-k: The way this function's docstring is written, it makes me expect that these two tickets are for enabling the the option, rather than a more recent change to disable it from the original state in RHEL 8.  I think everything would be clearer if the docstring described the desired state for all cases.  For example:

```
Checks the status of the "iommu.strict=0" kernel option on RHEL ARM AMIs. The option should be:
- Present on RHEL 8 images (RHBZ#1836058).
- Not present on RHEL 9 and newer images (CLOUDX-1517, RHEL-115405).
```

test_suite/generic/test_generic.py
```diff
@@ -806,6 +806,15 @@ def test_yum_group_install(self, host, instance_data):
...
             pytest.skip('OCI image-builder builds do not include RHUI configuration')
 
+        def cleanup_dev_tools():
+            with host.sudo():
+                print('Cleaning up Development tools packages...')
+                cleanup_result = host.run('yum -y groupremove "Development tools"')
+                if cleanup_result.failed:
+                    print(f'Warning: Failed to clean up Development tools: {cleanup_result.stderr}')
+
+        request.addfinalizer(cleanup_dev_tools)
```
achilleas-k, line 815: I wonder what side-effects this can have for the image and tests that come later.  Is it possible it will remove packages that are needed in other tests?  I don't think it's a good idea, generally, to be modifying the system under test in ways that might affect the tests themselves.  If this step has any chance of removing a package that's required by (or checked for inclusion/exclusion by) a later test, then the order of tests becomes a determining factor for whether the tests will succeed or not.  Given that pytest makes no guarantees for test ordering, this will create random failures.  It might not be a problem now, but package groups change, either by the addition or removal of packages or indirect dependency changes.

When the issue came up last week about the inclusion of unexpected packages in the image, I thought the installation of this package group in particular was necessary.  Now I see in the docstring that this is just testing that `yum groupinstall` works.  Could this same test be handled by installing (and removing) a package with a few dependencies that would have a smaller chance of affecting system functionality?

If we really need to be installing a group to test these things, I'd recommend changing the cleanup to use [`dnf history`](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_software_with_the_dnf_tool/assembly_handling-package-management-history_managing-software-with-the-dnf-tool#reverting-dnf-transactions_assembly_handling-package-management-history) instead.

Also, probably better to use `dnf` instead of `yum` in general; I realise a lot of older users still type `yum` for package management out of habit, but if we're not testing any RHEL 7 images (do we test RHEL 7 images?), maybe it's time we made the switch.

achilleas-k (commented):
Thanks.  Some comments below.

test_suite/generic/test_generic.py
sshmulev: Good point!
I've updated the cleanup to use `dnf history undo last` which is more precise - it only removes packages installed in our test transaction, ensuring we don't modify the base system in unexpected ways.
Also switched from `yum` to `dnf` as suggested.

achilleas-k (dismissed):
Thanks.  LGTM.

achilleas-k (approved):

Thilak-KN (approved):
Approving since the primary reviewer approved other changes

PR #570, opened 2026-07-10
Proposal: Separate cloud-image-val utility from test suites

F-X64:
**Summary**
Proposes refactoring cloud-image-val from a monolithic Python test framework into a
minimal CLI tool focused solely on infrastructure orchestration. Tests would be
maintained independently by downstream teams in their own repositories.

**Problem**
The current codebase tightly couples infrastructure code (multi-cloud provisioning,
test execution) with test suites (RHEL/Fedora validation tests). This creates:
- Maintenance burden: Tests and utility can't evolve independently
- Coordination overhead: Image definition changes require coordinated PRs across repos
- Migration blockers: Prevents package testing teams from moving to Konflux
- Flexibility limits: Teams can't use their preferred testing frameworks

 **Proposed Solution**
Transform cloud-image-val into infrastructure-only:
- CIV responsibilities: Provision instances -> execute arbitrary test commands -> aggregate JUnit XML results
- Test teams responsibilities: Maintain tests in own repos using any framework -> produce JUnit XML output

This ADR documents the architectural vision, current pain points, proposed API contracts, and migration strategy.

Review Focus
- Does this separation of concerns align with our goals?
- Are the proposed API contracts (instance metadata, result format) sufficient?
- Is the migration strategy realistic?
Note: This is a proposal document (ADR-001) for discussion, not a finished guide.

### DIFF
--- a/docs/ADR-001-separation-of-concerns.md
+++ b/docs/ADR-001-separation-of-concerns.md
@@ -1,0 +1,609 @@
+# ADR: Separation of Cloud-Image-Val Utility from Test Suites
+
+**Status:** Proposed
+**Date:** 2026-07-10
+
+---
+
+## 1. Executive Summary
+
+This ADR proposes refactoring cloud-image-val (CIV) from a monolithic Python test framework into a minimal CLI tool that provisions cloud instances, executes arbitrary test commands, 
+and aggregates results into standardized reports. This approach enables downstream teams to maintain tests in their own repositories using any testing framework (pytest, shell scripts, Go tests, etc.) 
+while eliminating tight coupling and reducing CIVs maintenance burden.
+
+### Key Decisions
+1. CIV becomes infrastructure-only
+2. tests become completely independent
+3. CIV enforces standardized test input and result reporting (JUnit XML) to maintain CI/CD pipeline compatibility
+
+---
+
+## 2. Context
+
+The cloud-image-val (CIV) project currently combines two distinct concerns in a single monolithic repository:
+
+1. **The CIV Utility**: Multi-cloud infrastructure orchestration and test execution framework
+2. **Test Suites**: Cloud image validation tests for RHEL and Fedora across AWS, Azure, GCP (basic implementation), and OCI (in progress)
+
+This tight coupling creates significant maintenance burdens and prevents downstream teams from independently managing their test cases.
+
+### Current Code Problems
+
+**1. Tight Coupling Between Utility and Tests**
+- Test suites (`test_suite/`) import directly from utility code (`lib/`, `main/`)
+- Tests depend on environment variables set by `lib/config_lib.py` with no formal contract
+- Circular dependency risk: utility imports test runner, test runner imports utility libraries
+
+**2. Mixed Responsibilities**
+- `test_suite/conftest.py`: marker validation, JIRA integration, HTML report generation, test result modification
+- Test discovery logic hardcoded in `suite_runner.py` for cloud-specific test routing
+
+**3. External Dependencies Create Friction**
+- Hard dependency on `osbuild-composer-tests.rpm` for `/usr/libexec/tests/osbuild-composer/aws.sh`
+- Duplicated `ci/aws.sh` mirrors RPM functionality
+- Manual Terraform/OpenTofu version synchronization
+- `.gitlab-ci-cloud-experience.yaml` hardcodes runner names requiring manual updates
+
+**4. Prevents Independent Test Maintenance**
+- Image definitions live in osbuild-composer repository
+- Verification tests live in cloud-image-val repository
+- Chicken-and-egg problem: Can't test new image types without coordinated PRs across repositories
+- Package testing teams can't maintain tests independently
+- No way to version tests independently from infrastructure code
+
+**5. Code Quality Issues**
+- No type hints
+- Hardcoded magic numbers
+- Inconsistent naming (snake_case vs PascalCase inconsistently used)
+- No dependency separation (`requirements.txt` mixes test frameworks with runtime dependencies)
+- Security vulnerabilities (`os.system()` calls without sanitization)
+
+### Business Drivers
+
+1. Enable Konflux Migration: 
+- Migration is required long term
+- Package testing needs simpler pipeline without osbuild-composer dependency
+
+2. Improve Maintainability
+- GitLab pipelines too complex for package testing needs
+- Too much time spend fixing CIV bugs vs too little time spend developing CIV
+- PR pipeline state mostly ignored by devs
+  
+3. Better Logging
+- CIV doesn't provide package-testing-level detail
+
+4. Atomic Changes
+- Downstream teams need to commit image definitions + tests together
+
+5. Flexibility
+- Different teams have different testing needs (simple shell vs. complex pytest)
+
+---
+
+## 3. Decision
+
+Refactor CIV into a minimal infrastructure-only CLI tool that provisions cloud instances, executes arbitrary test commands, and aggregates results into standardized reports.
+
+**CIVs responsibilities:** 
+- Manage infrastructure lifecycle (provision → expose metadata → cleanup)
+- Execute arbitrary test commands
+- Aggregate and validate test results into standardized format (JUnit XML)
+- Provide consistent exit codes and reporting for CI/CD pipelines
+
+**Test teams responsibilities:** 
+- Write tests using any framework/language
+- Consume standardized instance metadata
+- Produce JUnit XML output (natively *or via CIV helpers*)
+
+---
+
+## 4. New Ownership Model
+<details open>
+  <summary>Ownership Model Diagrams</summary>
+
+### Current: Centralized
+
+```mermaid
+graph TB
+    subgraph "CloudX Owns Everything"
+        CIV_TEAM[CIV]
+        
+        INFRA[Infrastructure Code]
+        TESTS[All Test Suites]
+        CONFTEST[Test Framework]
+        HELPERS[Test Helpers]
+    end
+    
+    subgraph "Downstream Teams"
+        OSBUILD[osbuild Team]
+        CLOUDX[Package Maintainers]
+        QE["Cloud QE Team(s)"]
+    end
+    
+    CIV_TEAM --> INFRA
+    CIV_TEAM --> TESTS
+    CIV_TEAM --> CONFTEST
+    CIV_TEAM --> HELPERS
+    
+    OSBUILD -.->|request changes| CIV_TEAM
+    CLOUDX -.->|request changes| CIV_TEAM
+    QE -.->|request changes| CIV_TEAM
+```
+
+### Proposed: Distributed
+
+```mermaid
+graph TB
+    subgraph CloudX
+        CLOUDX_1[CLI Tool]
+        CLOUDX_2[Infrastructure Code]
+        CLOUDX_3[Test Helpers]
+    end
+    
+    subgraph osbuild["osbuild Team"]
+        OSBUILD_1[Image Definitions]
+        OSBUILD_2[Test Suites]
+        OSBUILD_3[Test Framework]
+    end
+    
+    subgraph packager["Package Maintainers"]
+        PACKAGER_1[Package]
+        PACKAGER_2[Test Suites]
+        PACKAGER_3[Test Framework]      
+    end
+    
+    subgraph cloud["Cloud Teams"]
+        CLOUD_1[Cloud Image]
+        CLOUD_2[Test Suites]        
+        CLOUD_3[Test Framework]
+    end
+    
+    osbuild -.->|runs| CloudX
+    packager -.->|runs| CloudX
+    cloud -.->|runs| CloudX
+
+    OSBUILD_2 -->|uses| OSBUILD_3
+    PACKAGER_2 -->|uses| PACKAGER_3
+    CLOUD_2 -->|uses| CLOUD_3
+
+    CLOUDX_1 -->|implements| CLOUDX_2
+    CLOUDX_1 -->|implements| CLOUDX_3
+```
+</details>
+
+---
+
+## 5. Proposed Architecture
+
+### Core Concept
+
+```
+┌────────────────────────────────────────────────────────────────────┐
+│  cloud-image-val                                                   │
+│  ┌────────────┐  ┌────────────┐  ┌────────────┐  ┌─────────────┐   │
+│  │ Provision  │→ │ Expose     │→ │ Execute    │→ │ Aggregate   │   │
+│  │ Instances  │  │ Metadata   │  │ Command    │  │ Results     │   │
+│  └────────────┘  └────────────┘  └────────────┘  └─────────────┘   │
+│                                                                    │
+│  Outputs:                                                          │
+│  - /tmp/civ-instances.json (IPs, username, cloud, distro)          │
+│  - /tmp/civ-ssh-config (SSH connection config)                     │
+│  - Environment variables (CIV_CLOUD, CIV_INSTANCES_COUNT)          │
+│  - Aggregated JUnit XML report (merged from all instances)         │
+│  - Normalized exit code (0=pass, 1=fail, 2=error)                  │
+└────────────────────────────────────────────────────────────────────┘
+                           │
+                           │ Executes any command (must produce JUnit XML)
+                           ▼
+        ┌──────────────────────────────────────────────────────┐
+        │  Downstream Test Repositories                        │
+        │  (These are just examples)                           │
+        ├──────────────────────────────────────────────────────┤
+        │  Option 1: Pytest                                    │
+        │  $ pytest tests/ --junit-xml=/tmp/results.xml        │
+        │                                                      │
+        │  Option 2: Shell script                              │
+        │  $ ./smoke-test.sh  # requires helper                │
+        │                                                      │
+        │  Option 3: Go tests                                  │
+        │  $ go test -json ./... > /tmp/results.json           │
+        │                                                      │
+        │  Option n: Custom binary (any framework)             │
+        │  $ ./my-validator --output-junit /tmp/results.xml    │
+        └──────────────────────────────────────────────────────┘
+                           │
+                           ▼
+                    Standard JUnit XML
+                (GitLab/Jenkins/Konflux compatible)
+```
+
+<details>
+<summary>Detailed Architecture Diagram</summary>
+
+```mermaid
+flowchart TB
+    %% ============================================================
+    %% LAYER 1: Core Orchestration
+    %% ============================================================
+    subgraph CORE["Core Cloud-Image-Val Components"]
+        direction TB
+            CLI["<b>cli.py</b><br/>Command Line Interface<br/>Entry Point"]
+        PROV["<b>provisioner.py</b><br/>Infrastructure Provisioning"]
+        META["<b>metadata.py</b><br/>Instance Metadata Handler<br/><i>Writes: [A] [B]</i>"]
+        EXEC["<b>executor.py</b><br/>Test Execution Controller<br/><i>Spawns: [D]</i>"]
+        RESULTS["<b>results.py</b><br/>Results Collection & Processing<br/><i>Reads: [C]</i>"]
+
+        subgraph INFRA_LAYER["Infrastructure Layer"]
+            direction LR
+            CONFIG["config.py"]
+            TF_CTRL["OpenTofu<br/>IaC Engine"]
+            SSH["ssh/client.py"]
+        end
+
+        subgraph RESULTS_LAYER["Results Processing Layer"]
+            direction LR
+            CONVERT["converters"]
+            XSD["junit.xsd<br/>Schema Validator"]
+        end
+    end
+
+    %% ============================================================
+    %% LAYER 2: Filesystem Contract (Shared State)
+    %% ============================================================
+    subgraph CONTRACT["Filesystem Contract Layer<br/><i>Shared state between cloud-image-val and downstream tests</i>"]
+        direction TB
+        INST_JSON["<b>[A]</b><br><b>civ-instances.json</b><br/>Instance metadata, IPs, cloud details"]
+        SSH_CFG["<b>[B]</b><br><b>civ-ssh-config</b><br/>SSH connection configuration"]
+        RESULTS_XML["<b>[C]</b><br><b>results.xml</b><br/>JUnit XML test results"]
+    end
+
+    %% ============================================================
+    %% LAYER 3: Downstream Test Suites
+    %% ============================================================
+    subgraph DOWNSTREAM["<b>[D]</b> Downstream Test Repositories<br/><i>Each test suite reads [A][B] and writes [C]</i>"]
+        direction LR
+        OSBUILD["<b>test_rhel9.py</b><br/>Python • osbuild-composer<br/><i>Optional: [E]</i>"]
+        PKG["<b>test-otel.sh</b><br/>Shell • cloudx-packages<br/><i>Optional: [E]</i>"]
+        GO["<b>cloud_test.go</b><br/>Go • integration tests"]
+    end
+
+    %% ============================================================
+    %% LAYER 4: Optional Helpers
+    %% ============================================================
+    HELPERS["<b>[E]</b> <b>Optional Helpers</b><br/>junit-reporter.*<br/>Test result formatting utilities"]
+
+    %% ============================================================
+    %% PRIMARY WORKFLOW: Sequential phases (1-4) - ONLY MAIN FLOW
+    %% ============================================================
+    CLI ==>|"<b>1.</b> provision"| PROV
+    PROV --> INFRA_LAYER
+    PROV ==>|"<b>2.</b> write files"| META
+    CLI ==>|"<b>3.</b> execute"| EXEC
+    CLI ==>|"<b>4.</b> collect"| RESULTS
+    RESULTS --> RESULTS_LAYER
+
+    %% ============================================================
+    %% STYLING FOR VISUAL HIERARCHY
+    %% ============================================================
+
+    %% Entry point - strongest emphasis
+    style CLI stroke:#2C7A9B,stroke:#1a5a7a,stroke-width:4px,color:#ffffff
+
+    %% Core orchestration - primary color scheme
+    style CORE stroke:#E8F4F8,stroke:#2C7A9B,stroke-width:3px
+    style PROV stroke:#5A9FBF,stroke:#2C7A9B,stroke-width:2px,color:#ffffff
+    style META stroke:#5A9FBF,stroke:#2C7A9B,stroke-width:2px,color:#ffffff
+    style EXEC stroke:#5A9FBF,stroke:#2C7A9B,stroke-width:2px,color:#ffffff
+    style RESULTS stroke:#5A9FBF,stroke:#2C7A9B,stroke-width:2px,color:#ffffff
+
+    %% Infrastructure layer - supporting components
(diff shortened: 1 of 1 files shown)


docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,626 @@
...
+**Status:** Proposed
+**Date:** 2026-07-10
+
+---
+
+## 1. Executive Summary
+
+This ADR proposes refactoring cloud-image-val (CIV) from a monolithic Python test framework into a minimal CLI tool that provisions cloud instances, executes arbitrary test commands, 
+and aggregates results into standardized reports. This approach enables downstream teams to maintain tests in their own repositories using any testing framework (pytest, shell scripts, Go tests, etc.) 
+while eliminating tight coupling and reducing CIVs maintenance burden.
```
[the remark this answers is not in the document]
cmchase: Nit-picky, but accurate

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+- `test_suite/conftest.py`: marker validation, JIRA integration, HTML report generation, test result modification
+- Test discovery logic hardcoded in `suite_runner.py` for cloud-specific test routing
+
+**3. External Dependencies Create Friction**
+- Hard dependency on `osbuild-composer-tests.rpm` for `/usr/libexec/tests/osbuild-composer/aws.sh`
+- Duplicated `ci/aws.sh` mirrors RPM functionality
+- Manual Terraform/OpenTofu version synchronization
+- `.gitlab-ci-cloud-experience.yaml` hardcodes runner names requiring manual updates
+
+**4. Prevents Independent Test Maintenance**
```
cmchase, line 47: Nit: I think it's fair to call out that we have both a maintenance _and_ an ownership problem. CIV is used across several orgs within RH and the tight coupling creates bottlenecks for moving quickly.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+- Image definitions live in osbuild-composer repository
+- Verification tests live in cloud-image-val repository
```
cmchase, line 49: Suggestion: these two read a little oddly with the statement that they are the chicken-and-egg, but the point remains that we're coupling dependency and consumer concerns.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+- No way to version tests independently from infrastructure code
+
+**5. Code Quality Issues**
+- No type hints
+- Hardcoded magic numbers
+- Inconsistent naming (snake_case vs PascalCase inconsistently used)
+- No dependency separation (`requirements.txt` mixes test frameworks with runtime dependencies)
+- Security vulnerabilities (`os.system()` calls without sanitization)
+
+### Business Drivers
```
cmchase, line 61: Suggestion: we should consider adding speed to market as a business driver. The mandate from leaders is pretty clear on the ramp up on CVEs and it makes sense that we'll introduce more image testing over time and the coupling is both a slowdown and a risk to stability.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+
+1. Enable Konflux Migration: 
+- Migration is required long term
+- Package testing needs simpler pipeline without osbuild-composer dependency
+
+2. Improve Maintainability
+- GitLab pipelines too complex for package testing needs
+- Too much time spend fixing CIV bugs vs too little time spend developing CIV
+- PR pipeline state mostly ignored by devs
```
cmchase, line 70: Non-blocking Question: is the pipeline ignored by test contributors or developers in general?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+            direction LR
+            CONVERT["converters"]
+            XSD["junit.xsd<br/>Schema Validator"]
+        end
+    end
+
+    %% ============================================================
+    %% LAYER 2: Filesystem Contract (Shared State)
+    %% ============================================================
+    subgraph CONTRACT["Filesystem Contract Layer<br/><i>Shared state between cloud-image-val and downstream tests</i>"]
```
cmchase, line 253: Nit Question: will CIV define and expose this contract to consumers or is this something already exposed in the FS?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+The architecture relies on defined JSON and XML files as the integration contract between cloud-image-val and downstream test repositories. This loose coupling allows for language independent test suites (Python, Shell, Go).
+
+**Reference Labels:**
+- `[A]` = civ-instances.json (read by all test suites)
+- `[B]` = civ-ssh-config (read by all test suites)
+- `[C]` = results.xml (written by all test suites, read by results.py)
+- `[D]` = Downstream test repositories (spawned by executor.py)
+- `[E]` = Optional helper utilities (junit-reporter.*)
+
+**Decoupled Design:**
```
cmchase, line 339: Praise: I really like the approach. It allows for team optionality.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+```bash
+export CIV_CLOUD=aws
+export CIV_INSTANCES_JSON=/tmp/civ-instances.json
+export CIV_SSH_CONFIG=/tmp/civ-ssh-config
+export CIV_INSTANCES_COUNT=3
+```
+
+### CIV Helper Libraries
+
+To make result reporting easy for all test frameworks, CIV needs to provides helper libraries for languages that do not provide native JUnit output or auto conversion.
```
cmchase, line 397: Non-blocking question: is there a generic standard for conversion or would CIV need to develop these for each language separately?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+   - Uses `os.system(f'rm -f {self.config_path}')` for file deletion
+   - **Fix:** Replace with `os.remove()` with appropriate exception handling
+
+### Architectural Smells
+1. **Mixed Responsibilities** (`conftest.py`)
+   - Marker validation
+   - JIRA API integration
+   - HTML report generation
+   - Test result modification
+   - **Fix:** split into plugins
```
cmchase, line 470: Non-blocking Question: perhaps too implementation-specific for now, but would these plugins live in the CIV repo or would they get kicked out to a utility library/

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+1. **Magic Numbers**
+   ```python
+   max_processes = 162  # Why 162? 
+   max_reruns = 3       # Hardcoded retry logic
+   rerun_delay_sec = 5  # No justification
+   ```
+
+2. **Inconsistent Naming Conventions**
+   - Python: `AzureConfigBuilderV2` vs `aws_config_builder`
+   - **Fix:** Use flake8 naming consistently in CI (future cleanup)
```
cmchase, line 492: Non-blocking, implementation question: would we run these as pre-commits or would they just live on CI?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+   - **Fix:** Remove dependency after osbuild migration
+
+---
+
+## 7. Risks
+
+| Risk | Impact | Solution |
+|------|--------|------------|
+| Breaking existing CI pipelines | HIGH | upgrade in staging environment first |
+| Invalid JUnit XML output | HIGH | Mandatory XSD validation before CIV returns. Clear error messages guide test authors |
```
cmchase, line 527: Non-blocking, implementation question: what would the feedback loop look like for test authors? Could they run this on their local to quickly iterate or would pipelines be involved? (Sorry for the newbie question, I'm not familiar with the dev experience.)

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+| Large result files | LOW | Add optional compression for result files. We can still implement streaming XML merge if files exceed a threshold |
+
+---
+
+## 8. Success Criteria
+
+1. **CIV Core Simplification**: Core codebase reduced. Structure needs to guide developers.
+2. **Language Flexibility**: At least one spike using non-Python tests (shell or Go)
+3. **Downstream Adoption**: 2+ teams using new approach in production (osbuild, packages maintainers)
+4. **Performance**: Infrastructure provision time unchanged or faster (remove pytest overhead)
```
cmchase, line 543: Non-blocking question: do we have baseline metrics for this currently or do we need to capture them?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+5. **Security**: No bandit high-severity findings
+6. **Type Safety**: Aiming for 80% type hint coverage in `core/` and `ssh/`
```
cmchase, line 545: Non-blocking Question: are we looking to `warn` and not `error` on initial linting checks?

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+10. **Stretchgoal: Helper Library Coverage**:
+    - Optional: Shell helper covers 90%+ of common test patterns
+    - Optional: At least one pilot uses shell helper successfully (package testing)
+    - Optional: Go JSON converter tested with real output
+
+---
+
+## 9. Open Questions for Review
+
+1. Use `/tmp/civ-*.json` or configurable path?
```
cmchase, line 561: Observation: FWIW, Sourcery called this out in a previous comment.

cmchase (commented):
I think this is great overall and I really appreciate the time and thought you put into the initial design, but there are a lot of technical details I think we'll need input on from teams that consume CIV.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+└── README.md
+```
+To be removed:
+- test_suite/ (entire directory moves to separate repos)
+- lib/test_lib.py (tests don't import from CIV)
+- lib/aws_lib.py (moved to cloud/aws/)
+- lib/console_lib.py (merged into cli.py)
+
+
+### Optional Helper Package (Separate Repository)
```
sshmulev, line 436: My proposal for this idea is:
Each test domain folder (tests/network/, tests/storage/, etc.) contains its own fixtures.py and helpers.py for domain-specific stuff, and there's a top-level helpers/common/ for things shared across all domains.

Rather than organizing tests by cloud provider (test_aws.py, test_azure.py) or the current generic/ vs cloud/ split, I'd suggest organizing by test domain:


```
tests/
   **common/              # shared helpers/fixtures used across domains
       helpers.py
       fixtures.py**
   network/
       test_network.py
       helpers.py
       fixtures.py
   storage/
       test_storage.py
       helpers.py
       fixtures.py
   packages/
       test_packages.py
       helpers.py
       fixtures.py
   certificates/
       test_packages.py
       helpers.py
       fixtures.py
   security/
       test_packages.py
       helpers.py
       fixtures.py
   boot/
       test_packages.py
       helpers.py
       fixtures.py
     ...
```

Each test uses cloud markers (@pytest.mark.run_on(['aws', 'azure'])) to declare where it runs, we already have this pattern. Domain-specific fixtures and helpers live alongside the tests that need them, and truly shared utilities 
 (like AVC checks) go in common/.

docs/ADR-001-separation-of-concerns.md
F-X64: Right, the chicken and egg problem occurs once we try to change one of them without at the same changing the other (e.g. synced PRs). I can add a bit more information to this.

docs/ADR-001-separation-of-concerns.md
F-X64: Very true, the ownership model that is proposed a little further along in section 4 visualizes this.
Maybe we can add something like

```suggestion
**4. Prevents Independent Test Maintenance**
- Split repo ownership across teams/orgs blurs responsibilities, slowing down project decisions and steering
```

docs/ADR-001-separation-of-concerns.md
F-X64: Good thinking. Maybe something like thios?

```suggestion
### Business Drivers
...
6. Speed to Market
- Current system coupling creates a bottleneck, directly conflicting with our interest to rapidly address and resolve CVEs.
- As we introduce more image testing over time, this tight coupling will increasingly slow down our release cycles and threaten overall system stability.
```

docs/ADR-001-separation-of-concerns.md
F-X64: I'd say anyone opening a PR.
Schutzbot hasn't been green in a very long time even though we require it to pass to keep changes between projects (osbuild / CIV) in sync.

docs/ADR-001-separation-of-concerns.md
F-X64: I'd like to expose it.
I don't want to rely on any magically created files that are supposed to be in very specific locations on the host running the tests.
This is one of the issues I'd like to get feedback on from the teams implementing tests.

docs/ADR-001-separation-of-concerns.md
F-X64: The only standard we have in this case is that it has to be output in JUnit.
Most frameworks have existing exporters, e.g.:
- pytest --junit-xml=results.xml (native no additional changes necessary) 
- gotestsum --junitfile results.xml (wrapper)
- ./my-script.sh | tap-junit > path/to/results.xml (there are packages for the test anything protocol)

So in this sense, we didn't even need to develop anything. The helper should mainly cover edgecases.

docs/ADR-001-separation-of-concerns.md
F-X64: Those would live in the CIV repo. Maybe plugin is the wrong term here. We can consider these modules of the project.

docs/ADR-001-separation-of-concerns.md
F-X64: Pre-commit would be my go-to for flake8.
We should still run them on PRs anyways.

docs/ADR-001-separation-of-concerns.md
F-X64: Yes running this on your local machine would be no problem.
In fact, you wouldn't need CIV for local testing at all as tests and CIV wouldn't require each other. For running tests on test instances (e.g. on real cloud instances) the flow would look like this:

 1. Write tests using your own framework.
 2. Ensure that your tests consume the civ-instances.json file to connect to provisioned instances (file path can be configurable or not, depending on what we decide on).
 3. Ensure that your tests produce JUnit XML output
 4. CIV takes care of provisioning, running your tests, cleanup, aggregating results and outputting the final JUnit results.
 5. Once you're confident, add a CI job, ideally using the CIV container image.

docs/ADR-001-separation-of-concerns.md
F-X64: I would be okay with a warning level. Error seems a bit excessive.

docs/ADR-001-separation-of-concerns.md
F-X64: I believe @sshmulev did collect some.
We should have enough opportunity to run some performance tests though.

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+export CIV_SSH_CONFIG=/tmp/civ-ssh-config
+export CIV_INSTANCES_COUNT=3
+```
+
+### CIV Helper Libraries
+
+To make result reporting easy for all test frameworks, CIV needs to provides helper libraries for languages that do not provide native JUnit output or auto conversion.
+Alternative: CIV includes converters for common test output formats, e.g. Golang.
+
+### Simplified CIV Repository Structure
```
sshmulev, line 400: Rather than organizing tests by cloud provider (test_aws.py, test_azure.py) or the current generic/ vs cloud/ split, I'd suggest organizing by test domain:

```
tests/
   common/              # shared helpers/fixtures used across domains
       helpers.py
       fixtures.py
   network/
       test_network.py
       helpers.py
       fixtures.py
   storage/
       test_storage.py
       helpers.py
       fixtures.py
   packages/
       test_packages.py
       helpers.py
       fixtures.py
   certificates/
       test_certificates.py
       helpers.py
       fixtures.py
   security/
       test_security.py
       helpers.py
       fixtures.py
   boot/
       test_boot.py
       helpers.py
       fixtures.py
     ...
```
Each test uses cloud markers (@pytest.mark.run_on(['aws', 'azure'])) to declare where it runs. Domain-specific fixtures and helpers live alongside the tests that need them, and truly shared utilities (like AVC checks) go in common/.

On top of this, I think we should add tier markers to classify test priority:

Tier 1 - Core functionality, must always pass. A failure here is an immediate blocker (e.g., boot, SSH access, networking basics, correct image identity).
Tier 2 - Important regression checks, but a failure doesn't block the release immediately. Needs investigation on a short timeline
Tier 3 - Nice-to-have validations, edge cases, non-critical regressions.

Usage would look like:
   @pytest.mark.tier1
   @pytest.mark.run_on(['all'])
   def test_sshd(self, host):
       ...
  
   @pytest.mark.tier2
   @pytest.mark.run_on(['aws', 'azure'])
   def test_yum_group_install(self, host):
       ...
Benefits:

Combined with domain folders, this gives you two axes of triage: "5 network failures, 3 are tier-1" is immediately actionable without digging through logs. It also lets CI pipelines gate on tier-1 only for fast feedback, while running tier-2/3 in parallel or post-merge.

Ownership by expertise: domain-specific fixtures and helpers live next to the tests that use them, and shared ones go in helpers/common/.

Current test_generic.py is already implicitly domain-grouped , there are clusters of network tests (test_eth0_network_adapter_setup, test_network_manager_cloud_setup, test_network_manager_conf_plugins), package tests
(test_required_packages_are_installed, test_yum_plugins, test_yum_package_install), security tests (test_selinux_mode, test_no_avc_denials, test_auditd), etc. This would formalize that grouping.

Cloud-specific tests (like AWS identity checks or Azure SRIOV rules) can still live in their domain folder with a cloud marker, or in a cloud_specific/ folder if they're truly provider-only

docs/ADR-001-separation-of-concerns.md
cmchase: LGTM

docs/ADR-001-separation-of-concerns.md
cmchase: LGTM

docs/ADR-001-separation-of-concerns.md
F-X64: There might be a misunderstanding of the intended purpose of this helper package.
It's not supposed to provide helpers for writing tests, its only purpose would be to provide some sort of export functionality to make it easier for developer teams to write tests that can keep the contract between CIV and their test input / output. E.g. something that makes it super easy for none python languages to export results in JUnit.

docs/ADR-001-separation-of-concerns.md
F-X64: I assume you're proposing a new structure for the existing tests? 
Yeah that would work. 

We need to clarify first whether any of the existing tests will even continue to be maintained by us or whether all of this ownership moves to another team. If we keep maintaining these test cases in a separate repository we can do this.  I wouldn't force a specific way of organizing tests on the development teams though as long as they keep the contract between CIV and their tests.

achilleas-k:
I had a quick read through and it looks good on first pass.  I'm gone until next week Tuesday (back on the 21st), so I didn't want to leave this unread until I'm back, but I don't have time to do a nitpicky review now.  So I'm not approving the PR, but overall, this looks very sane to me.

Thanks!

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+├── cli.py                    # CLI entry point (refactored from cloud-image-val.py)
+├── requirements.txt          # MINIMAL: PyYAML, paramiko, sshconf, requests, lxml
+├── pyproject.toml            # Optional: Package for pip install (convenience)
+└── README.md
+```
+To be removed:
+- test_suite/ (entire directory moves to separate repos)
+- lib/test_lib.py (tests don't import from CIV)
+- lib/aws_lib.py (moved to cloud/aws/)
+- lib/console_lib.py (merged into cli.py)
```
knivnia, line 433: line 579 says it'll move to `console.py`

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+---
+
+## Appendix: File Changes
+
+### CIV Repository - Files to Refactor
+- `cloud-image-val.py` → `cli.py` (simplified CLI)
+- `main/cloud_image_validator.py` → `core/provisioner.py` (remove test execution)
+- `lib/config_lib.py` → `core/config.py` (cleanup, type hints)
+- `lib/ssh_lib.py` → `ssh/client.py`
+- `lib/console_lib.py` → `utils/console.py`
```
knivnia, line 579: line 433 says it'll move to `cli.py`

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+   ```
+   - **Fix:** New executor returns actual exit code from subprocess.run()
+
+6. **Duplicate Code**
+   - `osbuild-composer RPM script` duplicates and, in some parts, modifies `ci/aws.sh`
+   - **Fix:** Remove after osbuild-composer migration
+
+### Dependency Issues
+1. **No Separation of Test vs Runtime Dependencies**
+   - 9/18 packages in `requirements.txt` are test-only
```
knivnia, line 514:
```suggestion
   - 9/17 packages in `requirements.txt` are test-only
```
typo 😛

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+## 2. Context
+
+The cloud-image-val (CIV) project currently combines two distinct concerns in a single monolithic repository:
+
+1. **The CIV Utility**: Multi-cloud infrastructure orchestration and test execution framework
+2. **Test Suites**: Cloud image validation tests for RHEL and Fedora across AWS, Azure, GCP (basic implementation), and OCI (in progress)
+
+This tight coupling creates significant maintenance burdens and prevents downstream teams from independently managing their test cases.
+
+### Current Code Problems
```
knivnia, line 30: this part intersects with Section 6, should we consolidate them? Or shorten this part as sort of a 'motivation for change' and leave details for section 6

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+    - Optional: At least one pilot uses shell helper successfully (package testing)
+    - Optional: Go JSON converter tested with real output
+
+---
+
+## 9. Open Questions for Review
+
+1. Use `/tmp/civ-*.json` or configurable path?
+2. Should CIV provision instances sequentially or in parallel? (Affects cloud rate limits)
+3. fail hard on invalid JUnit XML (exit code 2) or warn and continue with raw output?
```
knivnia, line 563: imo, if we have 

> CIV enforces standardized test input and result reporting (JUnit XML) to maintain CI/CD pipeline compatibility

then I'd say it should be hard fail, otherwise what's the point? Also, that way it will be separated from linting warnings

knivnia (commented):
Should we discuss and add a section about migration plan/order of operations? So other teams be more prepared for construction works?

docs/ADR-001-separation-of-concerns.md
F-X64: Good catch! console.py was based on an older draft. Messed that up in copy/pasting.
So cli.py is correct. I also notice that I forgot to add cloud/aws/ into the proposed file structure..

docs/ADR-001-separation-of-concerns.md
```diff
@@ -0,0 +1,609 @@
...
+│   └── junit.xsd             # JUnit XML schema for validation
+├── cli.py                    # CLI entry point (refactored from cloud-image-val.py)
+├── requirements.txt          # MINIMAL: PyYAML, paramiko, sshconf, requests, lxml
+├── pyproject.toml            # Optional: Package for pip install (convenience)
+└── README.md
+```
+To be removed:
+- test_suite/ (entire directory moves to separate repos)
+- lib/test_lib.py (tests don't import from CIV)
+- lib/aws_lib.py (moved to cloud/aws/)
```
F-X64, line 432:
```suggestion
- lib/aws_lib.py (integrate into cloud/...)
```

docs/ADR-001-separation-of-concerns.md
F-X64: I'm impressed

docs/ADR-001-separation-of-concerns.md
F-X64: Yep this is wrong, thanks for catching this!
Should not have added these changes in multiple places 

```suggestion
- `lib/console_lib.py` → `cli.py`
```

F-X64:
> Should we discuss and add a section about migration plan/order of operations? So other teams be more prepared for construction works?

Sounds good to me but I'd like to get overall agreement on whether we should proceed with this idea first.

src-up:
Reviewed this — thanks for the write up. 

- I tried sketching an example (say RHEL-on-AWS vs. RHEL-on-Azure) and kept landing on the same question: how do we avoid each downstream team re-implementing cloud-agnostic checks like "does SSH work"? 

- Is my understanding correct that CIV and the test command seem to require running on one host (via local /tmp/civ-* files) meaning users need to clone both repos (new CIV + new tests repo) onto that host?

F-X64:
Thanks for your review @src-up 

> * I tried sketching an example (say RHEL-on-AWS vs. RHEL-on-Azure) and kept landing on the same question: how do we avoid each downstream team re-implementing cloud-agnostic checks like "does SSH work"?

Yes it makes sense to have the baseline aka the generic test cases available per default. We can provide an option to run them through a flag like "--baseline" or "--baseline rhel9".

I don't want to keep them in the same place as the CIV tool as I am aiming for a complete split of tooling and test cases. Moving the existing tests to a separate repo that can be loaded in along side your more specific test cases would be possible. E.g. that would be the step mentioned in line 430:

> - test_suite/ (entire directory moves to separate repos)

That said, it would make sense to be able to reference multiple test sources (e.g. repositories) for one test execution.

> * Is my understanding correct that CIV and the test command seem to require running on one host (via local /tmp/civ-* files) meaning users need to clone both repos (new CIV + new tests repo) onto that host?

In short: No, the way CIV operates wouldn't change much. 

CIV will run in any execution environment which does not need to be the host you're testing. 
The orchestration and execution of tests will still happen through SSH.
In CI for example, the pipeline uses the CIV container as the job image, clones the user's repo (not into the host that runs the tests), and executes the tests through SSH on the hosts.

The flow would look like this:

1. Spawn the CIV container (or run it locally), clone the tests (suggestion: should we use a golang approach here and make CIV clone the repos?)
2. CIV provisions remote cloud instances
3. CIV writes metadata locally e.g. into /tmp/civ-instances.json and /tmp/civ-ssh-config
4. CIV executes the test subprocess command which uses the metadata files and SSH information to run the test on the remote instances.

cmchase (approved):
This has been open for several weeks and our calls for additional reviews haven't generated more feedback. Let's go forward with the directions proposed.

src-up:
Overall, I think this PR is in the right direction.

PR #586, opened 2026-07-27
Fix OCI tests for sshd config and network driver chec

psiayn:
- test_sshd_config: update sshd to check for uncommented PasswordAuthentication yes
- test_correct_network_driver_is_used: update network driver to check the correct path


sshmulev (approved):
LGTM+1

PR #597, opened 2026-08-12
Add JSON Schema for CIV instance metadata contract (CIV-005)

sshmulev:
Formalizes the implicit instance data format written by cloud_image_validator.py as a JSON Schema, enabling downstream consumers to validate and document the expected fields.

[CLOUDX-2046](https://redhat.atlassian.net/browse/CLOUDX-2046)

[CLOUDX-2046]: https://redhat.atlassian.net/browse/CLOUDX-2046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schemas/civ-instances.schema.json
+++ b/schemas/civ-instances.schema.json
@@ -1,0 +1,69 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "CIV Instance Metadata Contract",
+    "description": "Defines the integration contract between CIV provisioner and downstream test suites. The top-level object is a map of instance identifiers (e.g. OpenTofu resource addresses) to instance metadata objects.",
+    "type": "object",
+    "additionalProperties": {
+        "$ref": "#/$defs/instance"
+    },
+    "$defs": {
+        "instance": {
+            "type": "object",
+            "description": "Metadata for a single provisioned cloud instance.",
+            "required": [
+                "name",
+                "address",
+                "username",
+                "cloud",
+                "image"
+            ],
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "Human-readable instance name, typically derived from the OpenTofu resource name."
+                },
+                "address": {
+                    "type": "string",
+                    "description": "Network address used for SSH connections. Resolved from public_dns, public_ip, or private_ip in order of preference."
+                },
+                "username": {
+                    "type": "string",
+                    "description": "SSH login username for the instance (e.g. 'ec2-user', 'cloud-user')."
+                },
+                "cloud": {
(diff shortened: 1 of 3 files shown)


schemas/civ-instances.schema.json
```diff
@@ -0,0 +1,69 @@
...
+                "distro": {
+                    "type": "string",
+                    "description": "Linux distribution identifier (e.g. 'rhel', 'fedora')."
+                },
+                "version": {
+                    "type": "string",
+                    "description": "Distribution version (e.g. '9.3')."
+                },
+                "arch": {
+                    "type": "string",
```
F-X64, line 54: Suggestion: String is completely fine but arch could be and enum e.g. ["x86_64", "aarch64"] like the cloud field.

schemas/civ-instances.schema.json
```diff
@@ -0,0 +1,69 @@
...
+                "image_id": {
+                    "type": "string",
+                    "description": "Cloud-specific image identifier (e.g. AMI ID, image URN)."
+                },
+                "instance_type": {
+                    "type": "string",
+                    "description": "Cloud-specific instance/VM size (e.g. 't3.medium', 'Standard_D2s_v3')."
+                }
+            },
+            "additionalProperties": true
```
F-X64, line 67: Nice, OpenTofu produces some provider specifics that aren't in the schema so this is the right call.

schemas/civ-instances.schema.json
```diff
@@ -0,0 +1,69 @@
...
+                    "description": "Network address used for SSH connections. Resolved from public_dns, public_ip, or private_ip in order of preference."
+                },
+                "username": {
+                    "type": "string",
+                    "description": "SSH login username for the instance (e.g. 'ec2-user', 'cloud-user')."
+                },
+                "cloud": {
+                    "type": "string",
+                    "description": "Cloud provider identifier.",
+                    "enum": ["aws", "azure", "gcloud", "oci"]
```
F-X64, line 36: I like that you've used gcloud not gcp.
Let's use this consistently across all implementations. We have been mixing them in the past.

schemas/civ-instances.schema.json
```diff
@@ -0,0 +1,69 @@
...
+                },
+                "image": {
```
F-X64, line 38: Good call on not adding a type restrain here.

F-X64 (approved):
Solid implementation. The schema is well designed! 
Whether you stick with arch as a string or change to enum is up to you.

PR #598, opened 2026-08-14
CIV-007: Add instance metadata writer

F-X64:
**Summary**
Added instance metadata writer with typed dataclass and schema valdation.

**Details**
Consolidate instance metadata writing, SSH config generation, and CIV environment variable export into a single module (core/metadata.py) with typed fields, input validation, and JSON Schema enforcement as implemented in CIV-005.

Added test coverage for AWS, Azure, GCloud and OCI instance type metadata.
There is still an issue with the way instance metadata is handled as long as we use the instance address as an identifier but the chances of overlap are basically zero as long as CIV runs are dealing with one cloud provider at a time.
Still, I've added a comment for clarification.

A word about the unit tests:
I'm using monkeypatch to avoid polluting the test space with the necessary envs.

### DIFF
--- a/core/metadata.py
+++ b/core/metadata.py
@@ -1,0 +1,130 @@
+from __future__ import annotations
+
+import functools
+import json
+import os
+from dataclasses import asdict, dataclass, fields
+from pathlib import Path
+from typing import Any, Union
+
+import jsonschema
+import sshconf
+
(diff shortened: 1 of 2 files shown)


sshmulev (approved):
Everything looks good, Clean dataclass design, good schema enforcement, and thorough test coverage across all four cloud providers.
LGTM+1

PR #600, opened 2026-08-17
Civ 006 junit xsd schema

sshmulev:
## Summary
  - Adds `schemas/junit.xsd`: JUnit XML Schema Definition based on the Windyroad standard, relaxed to accept `pytest --junit-xml` output
  - Covers standard elements: `testsuites`, `testsuite`, `testcase`, `failure`, `error`, `skipped`, `system-out`, `system-err`
  - Adds `lxml==5.3.1` and `jsonschema==4.23.0` to `requirements.txt`
  - 8 unit tests: valid XML (pass/fail/skip/error/system-output), empty suite, rejection of malformed XML, and live pytest output validation

  ## Test plan
  - [x] XSD validates correct JUnit XML samples
  - [x] XSD rejects malformed XML (missing required attributes)
  - [x] Live `pytest --junit-xml` output passes validation

[  CLOUDX-2047
](https://redhat.atlassian.net/browse/CLOUDX-2047)

### DIFF
--- a/schemas/junit.xsd
+++ b/schemas/junit.xsd
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  JUnit XML Schema Definition for CIV result validation.
+
+  Based on the widely-used Windyroad JUnit Schema
+  (https://github.com/windyroad/JUnit-Schema) with relaxed constraints
+  to accept output from pytest - -junit-xml, which omits some elements
+  and attributes that the strict schema requires.
+
+  Copyright (c) 2011, Windy Road Technology Pty. Limited
+  Licensed under the Apache License Version 2.0
+-->
+<xs:schema
+    xmlns:xs="http://www.w3.org/2001/XMLSchema"
+    elementFormDefault="qualified"
+    attributeFormDefault="unqualified">
+
+    <xs:element name="testsuites">
+        <xs:annotation>
+            <xs:documentation xml:lang="en">Contains an aggregation of testsuite results</xs:documentation>
+        </xs:annotation>
+        <xs:complexType>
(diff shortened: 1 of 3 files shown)


test/test_schema_validation.py
F-X64, line 1: Very good test coverage.
Nit pick: Maybe add some negative tests for unknown or extra elements to round the test suite out.

schemas/junit.xsd
```diff
@@ -0,0 +1,131 @@
...
+                                    <xs:attribute name="name" type="xs:string" use="required"/>
+                                    <xs:attribute name="value" type="xs:string" use="required"/>
+                                </xs:complexType>
+                            </xs:element>
+                        </xs:sequence>
+                    </xs:complexType>
+                </xs:element>
+                <xs:element name="testcase" minOccurs="0" maxOccurs="unbounded">
+                    <xs:complexType>
+                        <xs:sequence>
```
F-X64, line 58: Using xs:sequence means that the child elements (skipped, error, failure, system-out, system-err) need to be in exactly the order you've set them in. If a cli tool would return the exact elements but in a different order it would result in a validation failure. Honestly I don't know if this is worth fixing. Maybe we can just keep this in mind and if we thing this is an issue we can still change this to an xs:all or similar.

F-X64 (approved):
Looks good! Very solid implementation. I only have two nitpicks and I leave it open to you if you feel like they are worth addressing.

test/test_schema_validation.py
sshmulev: ACK

schemas/junit.xsd
sshmulev: ACK

PR #601, opened 2026-08-18
Civ 008 junit validation aggregation

sshmulev:
## Summary
  - Adds `core/results.py` with JUnit XML validation, multi-file merge, and exit code logic
  - `validate_junit_xml()` validates against `schemas/junit.xsd`, raises `ResultValidationError` with descriptive messages
  - `merge_results()` merges N JUnit XML files into one `<testsuites>` document, prefixing suite names with instance labels
  - `get_exit_code()` returns `0` (all pass), `1` (failures), or `2` (errors) per ADR convention
  - Handles edge cases: bare `<testsuite>` without wrapper, empty suites, mixed formats
  - 100% type hint coverage on all public APIs

  ## Test plan
  - [x] Valid single-file validation
  - [x] Multiple files merged with correct counter aggregation
  - [x] Instance labels prefix testsuite names
  - [x] Invalid/malformed XML raises `ResultValidationError`
  - [x] Empty result (0 test cases)
  - [x] Bare `<testsuite>` without `<testsuites>` wrapper
  - [x] Exit code 0/1/2 for pass/failure/error scenarios
  - [x] Errors take precedence over failures in exit code

[  CLOUDX-2049](https://redhat.atlassian.net/browse/CLOUDX-2049)

### DIFF
--- a/core/results.py
+++ b/core/results.py
@@ -1,0 +1,110 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from lxml import etree
+
+_XSD_PATH = Path(__file__).resolve().parent.parent / 'schemas' / 'junit.xsd'
+
+
+class ResultValidationError(Exception):
+    pass
+
+
+@dataclass
+class MergeResult:
+    total_tests: int
+    failures: int
+    errors: int
+    skipped: int
+    time: float
+
+
+def _load_xsd() -> etree.XMLSchema:
+    return etree.XMLSchema(etree.parse(str(_XSD_PATH)))
+
+
+def validate_junit_xml(path: str) -> bool:
+    try:
+        doc = etree.parse(path)
+    except etree.XMLSyntaxError as exc:
+        raise ResultValidationError(f"Malformed XML in {path}: {exc}") from exc
+
+    schema = _load_xsd()
+    if not schema.validate(doc):
+        errors = "; ".join(str(e) for e in schema.error_log)
+        raise ResultValidationError(f"XSD validation failed for {path}: {errors}")
+
+    return True
+
+
+def merge_results(
+    result_paths: list[str],
+    output_path: str,
+    instance_labels: list[str] | None = None,
+) -> MergeResult:
+    if instance_labels and len(instance_labels) != len(result_paths):
+        raise ValueError("instance_labels must match the length of result_paths")
+
+    schema = _load_xsd()
+    merged_root = etree.Element("testsuites")
+    total_tests = 0
+    total_failures = 0
+    total_errors = 0
+    total_skipped = 0
+    total_time = 0.0
+
+    for i, path in enumerate(result_paths):
+        try:
+            doc = etree.parse(path)
+        except etree.XMLSyntaxError as exc:
+            raise ResultValidationError(f"Malformed XML in {path}: {exc}") from exc
+
+        if not schema.validate(doc):
+            errors = "; ".join(str(e) for e in schema.error_log)
+            raise ResultValidationError(f"XSD validation failed for {path}: {errors}")
+
+        root = doc.getroot()
+        label = instance_labels[i] if instance_labels else None
+
+        if root.tag == "testsuites":
+            suites = root.findall("testsuite")
+        elif root.tag == "testsuite":
+            suites = [root]
+        else:
+            raise ResultValidationError(
+                f"Unexpected root element <{root.tag}> in {path}"
+            )
+
+        for suite in suites:
+            if label:
+                original_name = suite.get("name", "")
+                suite.set("name", f"{label}.{original_name}")
+
+            total_tests += int(suite.get("tests", 0))
+            total_failures += int(suite.get("failures", 0))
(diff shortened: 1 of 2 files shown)


core/results.py
```diff
@@ -0,0 +1,110 @@
...
+    return True
+
+
+def merge_results(
+    result_paths: list[str],
+    output_path: str,
+    instance_labels: list[str] | None = None,
+) -> MergeResult:
+    if instance_labels and len(instance_labels) != len(result_paths):
+        raise ValueError("instance_labels must match the length of result_paths")
```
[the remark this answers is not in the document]
F-X64: +1 on this. 
For context: instance_labels=[] is false in Python. If a caller e.g. passes an empty list with non empty results_paths no ValueError is raised and the empty labels are silently ignored.

core/results.py
```diff
@@ -0,0 +1,110 @@
...
+            doc = etree.parse(path)
+        except etree.XMLSyntaxError as exc:
+            raise ResultValidationError(f"Malformed XML in {path}: {exc}") from exc
+
+        if not schema.validate(doc):
+            errors = "; ".join(str(e) for e in schema.error_log)
+            raise ResultValidationError(f"XSD validation failed for {path}: {errors}")
+
+        root = doc.getroot()
+        label = instance_labels[i] if instance_labels else None
```
F-X64: Same as line 47 above: 
```suggestion
        label = instance_labels[i] if instance_labels is not None else None
```
By using "is not None" you are avoiding the issue.

core/results.py
```diff
@@ -0,0 +1,110 @@
...
+    time: float
+
+
+def _load_xsd() -> etree.XMLSchema:
+    return etree.XMLSchema(etree.parse(str(_XSD_PATH)))
+
+
+def validate_junit_xml(path: str) -> bool:
+    try:
+        doc = etree.parse(path)
```
F-X64: AFAIK there used to be a big vulnerability in the default parser for "external entity expansion" attacks. 
lxml / libxml2 solved this this in an update (need to look up which version) but whether this is safe is very much library verison dependent. 
To make this a little safer you can explicitly define a safe parser config at the top of the file and use it like this:

```
  _SAFE_PARSER = etree.XMLParser(
      resolve_entities=False, # This disables internal / external macro expansion, it's a little safer than just relying on the fixed lib being used
      no_network=True   # This makes surethat libxml2 cannot trigger network calls
  )
  
  ...
  etree.parse(path, _SAFE_PARSER)
```

F-X64: Ah there we go: https://security.snyk.io/vuln/SNYK-PYTHON-LXML-16119103
Just FYI

core/results.py
```diff
@@ -0,0 +1,110 @@
...
+    merged_root = etree.Element("testsuites")
+    total_tests = 0
+    total_failures = 0
+    total_errors = 0
+    total_skipped = 0
+    total_time = 0.0
+
+    for i, path in enumerate(result_paths):
+        try:
+            doc = etree.parse(path)
```
F-X64: Same as line 30. If you drop in the custom parser you're avoiding that vulnerability scenario.

core/results.py
```diff
@@ -0,0 +1,110 @@
...
+    skipped: int
+    time: float
+
+
+def _load_xsd() -> etree.XMLSchema:
+    return etree.XMLSchema(etree.parse(str(_XSD_PATH)))
+
+
+def validate_junit_xml(path: str) -> bool:
+    try:
```
F-X64, line 30: Hm, the logic in validate_junit_xml (29-37) and merge_results (59-66) which are implementing the XML parsing and validation blocks are nearly identical. At this point code duplication is perfectly fine but you could also extract the logic into a shared method like "_parse_and_validate(path, schema)" and call it from both. I'll leave that open to you.

core/results.py
```diff
@@ -0,0 +1,110 @@
...
+        doc = etree.parse(path)
+    except etree.XMLSyntaxError as exc:
+        raise ResultValidationError(f"Malformed XML in {path}: {exc}") from exc
+
+    schema = _load_xsd()
+    if not schema.validate(doc):
+        errors = "; ".join(str(e) for e in schema.error_log)
+        raise ResultValidationError(f"XSD validation failed for {path}: {errors}")
+
+    return True
```
F-X64: Tbh. this is exactly what the ticket said so you're spot on here.
THinking  about it, the -> bool return value makes no sense in this case as False is not a possible outcome. It either completes or raises an error.
A return type of -> None (raise only) would communicate the contract better. Minor issue as the implementation is accurate.

F-X64 (changes requested):
The module is very clean and well structured!
I have a couple of smaller comments, mainly on the parser config, every thing else is nit picking ;) 
Thanks for implementing this!

core/results.py
sshmulev: ACK

core/results.py
sshmulev: Good catch

F-X64 (approved):
Implementation looks great, thanks for adding the changes

PR #602, opened 2026-08-19
fix(oci): account for OCA and pqrpm GPG key dbs

src-up:
CIV counted only gpg-pubkey in the default rpmdb and expected 3 keys on all RHEL 9+.  Still not sure where that number 3 is from.

On OCI, Oracle Cloud Agent imports 2 Oracle Linux keys at first boot.

- On RHEL 9.7+ the Red Hat keys are imported by pqrpm into /usr/lib/pqrpm/lib/sysimage/rpm, so the default db only has the OCA keys.
- RHEL 10 keeps Red Hat keys in the default db (3 RH + 2 OCA).

So Expect 2 in the default db on OCI 9.7–9.x, 5 on OCI 10, and add test_pqrpm_gpg_keys for the 2 Red Hat keys in the pqrpm db. Other clouds are unchanged.

### DIFF
--- a/test_suite/generic/test_generic.py
+++ b/test_suite/generic/test_generic.py
@@ -647,16 +647,33 @@ class TestsGeneric:
     @pytest.mark.run_on(['all'])
     def test_number_gpg_keys(self, host, instance_data):
         """
-        Check that the number of GPGs is correct
+        Check that the number of GPGs is correct in the default rpmdb.
+
+        On OCI, Oracle Cloud Agent imports two Oracle Linux keys at first boot.
+        On RHEL 9.7+ OCI, Red Hat keys are in the pqrpm db (test_pqrpm_gpg_keys),
+        so this test expects only those 2 OCA keys in the default rpmdb.
+        On RHEL 10 OCI, Red Hat keys are in the default rpmdb (expect 5).
         """
+        release = version.parse(host.system_info.release)
+        is_oci = instance_data['cloud'] == 'oci'
+
         with host.sudo():
             # print the gpg public keys installed
             print(host.check_output('rpm -qa | grep gpg-pubkey'))
 
             if host.system_info.distribution == 'fedora':
                 num_of_gpg_keys = 1
+            elif is_oci and host.system_info.distribution == 'rhel' and \
+                    release >= version.parse('10'):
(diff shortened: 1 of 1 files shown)


test_suite/generic/test_generic.py
```diff
@@ -665,6 +682,33 @@ def test_number_gpg_keys(self, host, instance_data):
...
+
+        with host.sudo():
+            assert host.file(pqrpm_db).exists, \
+                f'pqrpm rpmdb missing at {pqrpm_db}'
+            print(host.check_output(
+                f'rpm --dbpath {pqrpm_db} -qa | grep gpg-pubkey'))
+
+        assert int(host.check_output(
+            f'rpm --dbpath {pqrpm_db} -q gpg-pubkey | wc -l')) == num_of_gpg_keys, \
+            f'There should be {num_of_gpg_keys} gpg key(s) installed in the pqrpm db'
```
sshmulev: You should shift this block into the "with host.sudo():" otherwisle it will fail.

test_suite/generic/test_generic.py
```diff
@@ -647,14 +647,31 @@ def test_pkg_signature_and_gpg_keys(self, host):
...
+                    release >= version.parse('10'):
+                # 3 Red Hat keys + 2 Oracle Cloud Agent keys
+                num_of_gpg_keys = 5
+            elif is_oci and host.system_info.distribution == 'rhel' and \
+                    version.parse('9.7') <= release < version.parse('10'):
+                # Red Hat keys are in the pqrpm db (test_pqrpm_gpg_keys);
+                # default rpmdb has the 2 Oracle Cloud Agent keys
+                num_of_gpg_keys = 2
             elif host.system_info.distribution == 'rhel' and \
                     version.parse(host.system_info.release) >= version.parse('9.0'):
```
sshmulev: redundant version.pars() call left  - need to change into 'release'

sshmulev (approved):
LGTM +1

src-up:
@F-X64 do  these changes look OK to you

F-X64:
@src-up Changes look great, I got nothing to add to @sshmulev 's review

F-X64 (approved):

PR #607, opened 2026-08-27
change test_yum_plugins to run only for 3p images

psiayn:
Adds an option `image_type` in resources.json which takes 1p or 3p.
If the image is 3p, only then the test `test_yum_plugins` runs. Otherwise it skips the test.


sshmulev (approved):
LGTM+1

PR #614, opened 2026-09-07
feat: add command executor with parallel SSH execution (CIV-012)

sshmulev:
## Summary
- Add `core/executor.py` with `execute()` function for running commands on cloud instances via SSH using `subprocess.run` with list args
- Uses `ThreadPoolExecutor` for parallel execution across instances and SCP for result file collection
- Typed `InstanceResult` and `ExecutionResult` dataclasses with `all_passed` property
- 14 unit tests with 100% coverage in `test/test_executor.py`

## Test plan
- [x] All 14 tests pass (`pytest test/test_executor.py -v`)
- [x] flake8 passes with no warnings
- [x] 100% code coverage verified

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -1,0 +1,116 @@
+from __future__ import annotations
+
+import os
+import subprocess
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from dataclasses import dataclass
+
+from core.metadata import InstanceMetadata
+
+
+@dataclass
+class InstanceResult:
+    instance_name: str
+    exit_code: int
+    result_file: str | None
+    stdout: str
+    stderr: str
+
+
+@dataclass
+class ExecutionResult:
+    instance_results: list[InstanceResult]
+
+    @property
+    def all_passed(self) -> bool:
+        return all(r.exit_code == 0 for r in self.instance_results)
+
+
+def _run_on_instance(
+    instance: InstanceMetadata,
+    command: str,
+    ssh_config: str,
+    results_dir: str,
+    remote_results_path: str,
+    timeout: int,
+) -> InstanceResult:
+    try:
+        result = subprocess.run(
+            [
+                "ssh",
+                "-F", ssh_config,
+                f"{instance.username}@{instance.address}",
+                command,
+            ],
+            capture_output=True,
+            text=True,
+            timeout=timeout,
+        )
+        exit_code = result.returncode
+        stdout = result.stdout
+        stderr = result.stderr
+    except subprocess.TimeoutExpired:
+        exit_code = 124
+        stdout = ""
+        stderr = f"SSH command timed out after {timeout} seconds"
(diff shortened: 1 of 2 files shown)


core/executor.py
```diff
@@ -0,0 +1,116 @@
...
+    result_file: str | None
+    stdout: str
+    stderr: str
+
+
+@dataclass
+class ExecutionResult:
+    instance_results: list[InstanceResult]
+
+    @property
```
F-X64, line 24: I like the addition of all_passed as a property, not a bool

core/executor.py
```diff
@@ -0,0 +1,116 @@
...
+                "-F", ssh_config,
+                f"{instance.username}@{instance.address}:{remote_results_path}",
+                local_result_file,
+            ],
+            capture_output=True,
+            timeout=60,
+        )
+        collected = scp_result.returncode == 0
+    except Exception:
+        collected = False
```
F-X64: When SSH fails, the code falls through to the SCP block and waits up to 60 seconds for an SCP connection to the same unreachable host. With parallel execution across a large number of instances, every unavailable instance adds 60 seconds of wait to the pipeline.

Let's track whether SSH itself failed at the connection level and skip SCP in that case:

```suggestion
def _run_on_instance(
    instance: InstanceMetadata,
    command: str,
    ssh_config: str,
    results_dir: str,
    remote_results_path: str,
    timeout: int,
) -> InstanceResult:
    ssh_failed = False
    try:
        result = subprocess.run(
            [
                "ssh",
                "-F", ssh_config,
                f"{instance.username}@{instance.address}",
                command,
            ],
            capture_output=True,
            text=True,
            timeout=timeout,
        )
        exit_code = result.returncode
        stdout = result.stdout
        stderr = result.stderr
    except subprocess.TimeoutExpired:
        ssh_failed = True
        exit_code = 124
        stdout = ""
        stderr = f"SSH command timed out after {timeout} seconds"
    except Exception as exc:
        ssh_failed = True
        exit_code = 1
        stdout = ""
        stderr = str(exc)

    local_result_file = os.path.join(
        results_dir, f"instance-{instance.name}.xml",
    )
    collected = False
    if not ssh_failed:
        try:
            scp_result = subprocess.run(
                [
                    "scp",
                    "-F", ssh_config,
                    f"{instance.username}@{instance.address}:{remote_results_path}",
                    local_result_file,
                ],
                capture_output=True,
                timeout=60,
            )
            collected = scp_result.returncode == 0
        except Exception:
            collected = False
```

F-X64 (changes requested):
Again, solid implementation! I'd recommend that we add the SCP fix in as it will safe significant time on SSH issues, other than that it looks good.  The sourcery comment about instance.name is valid, don't know if it is impacting our specific usecase though

F-X64 (approved):

PR #616, opened 2026-09-08
feat: add phase-based CLI entry point with subcommands (CIV-013)

sshmulev:
## Summary
- Add `cli.py` at repo root with 5 subcommands: `provision`, `execute`, `collect`, `run`, `cleanup`
- Maps to phase-based execution flow: each subcommand calls the appropriate `core/` module
- `civ run` ties the full pipeline (provision → execute → collect → cleanup) and accepts all flags from `cloud-image-val.py`
- `cloud-image-val.py` is not modified

## Dependencies
- Depends on CIV-011 (#613) and CIV-012 (#614) being merged first — imports `core.provisioner` and `core.executor`

## Test plan
- [x] 39 unit tests pass (`pytest test/test_cli.py -v`)
- [x] 98% code coverage
- [x] flake8 passes with no warnings
- [x] All 5 subcommands tested: parser, routing, and behavior
- [x] `cloud-image-val.py` unchanged

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -9,7 +9,7 @@ import sys
 
 from core.config import CoreConfig
 from core.executor import execute
-from core.metadata import InstanceMetadata
+from core.metadata import InstanceMetadata, write_instances_json
 from core.provisioner import Provisioner
 from core.results import get_exit_code, merge_results
 
@@ -39,7 +39,7 @@ def _build_config(args: argparse.Namespace) -> CoreConfig:
 
     for attr in ("debug", "parallel", "stop_cleanup"):
         val = getattr(args, attr, None)
-        if val:
+        if val is not None:
             config_dict[attr] = val
 
     if getattr(args, "tags", None):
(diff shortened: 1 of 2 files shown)


cli.py
```diff
@@ -0,0 +1,280 @@
...
+    p_coll = subparsers.add_parser(
+        "collect", help="Validate and merge JUnit XML results",
+    )
+    p_coll.add_argument("--results-dir", required=True,
+                        help="Directory containing XML result files")
+    p_coll.add_argument("-o", "--output-file", required=True,
+                        help="Output path for merged JUnit XML")
+    p_coll.set_defaults(func=cmd_collect)
+
+    # --- run ---
```
F-X64, line 218: Hm, this is missing  the "--attach" flag that exists in the legacy cloud-image-val.py.
In the refactor, "civ execute --instances-json" partially covers this (if pointed at existing instances), but the behavior is a bit different "--attach" also reuses .tf.json and .tfstate files. Is the attach flag in use at the moment or should we skip it?

cli.py
```diff
@@ -0,0 +1,280 @@
...
+    print(
+        f"Merged {merge_result.total_tests} tests: "
+        f"{merge_result.failures} failures, "
+        f"{merge_result.errors} errors, "
+        f"{merge_result.skipped} skipped"
+    )
+    return get_exit_code(merge_result)
+
+
+def cmd_run(args: argparse.Namespace) -> int:
```
F-X64, line 123: cmd_run (lines 120-161) reimplements the pipeline inline rather than calling cmd_provision, cmd_execute, and cmd_collect. This creates divergence risks.
E.g. the result collection logic diverges from cmd_collect: cmd_run uses !exec_result.instance_results[].result_file" (see lines 144 > ~146) while cmd_collect uses "glob.glob("*.xml") (see line 109)."

This does fulfill the task requirements but it might be solved more elegantly.

cli.py
```diff
@@ -0,0 +1,280 @@
...
+    if not resources_file:
+        raise SystemExit("error: --resources-file is required (or use --config-file)")
+
+    config_dict: dict[str, object] = {
+        "resources_file": resources_file,
+        "output_file": getattr(args, "output_file", None) or "/tmp/civ-results.xml",
+    }
+
+    for attr in ("debug", "parallel", "stop_cleanup"):
+        val = getattr(args, attr, None)
```
F-X64, line 41: The if val: check skips False values, which currenbtly works because CoreConfig defaults these to False.   If a future flag has default=True and the user doesn't pass it in it'll break. Same could happen if someone adds a boolean config key that defaults to True or the CoreConfig defaults change.
You can fix that by using "if val is not None:" instead of "if val:".

test/test_cli.py
```diff
@@ -0,0 +1,459 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+
+mock_provisioner_module = MagicMock()
+mock_executor_module = MagicMock()
+
+sys.modules.setdefault("core.provisioner", mock_provisioner_module)
+sys.modules.setdefault("core.executor", mock_executor_module)
```
F-X64, line 12: A heads-up: setdefault only inserts if the key is absent. Once core/provisioner.py (CIV-011) and core/executor.py (CIV-012) are merged, Python's import system will load the real modules, setdefault won't override them, and the cli import will try to import actual Provisioner/execute symbols. If those modules have unresolvable imports of their own (OpenTofu, paramiko, etc.), every test in this file will fail at import time. Or in other owrds, lets get the other tasks merged and see what needs fixing-

F-X64 (changes requested):
Already looking good! There are some issues but we need to merge CIV-011 and CIV-012 first.

F-X64 (approved):

